PR #85, opened 2018-10-24
Fix issue with naming when no subscription is present in context

cormacpayne:
Fix for https://github.com/Azure/azure-powershell/issues/7453


cormacpayne:
Sign job: https://azuresdkci.westus2.cloudapp.azure.com/job/ps-common-sign/199/

PR #92, opened 2018-11-05
Update profile models for ps1xml table format

maddieclayton:


maddieclayton:
@markcowl Example output can be found here: https://github.com/Azure/azure-powershell/pull/7755

PR #93, opened 2018-11-05
Change namespace for types in types.ps1xml

maddieclayton:


src/Authentication.Abstractions/AzureRmProfileProvider.cs
```diff
@@ -13,6 +13,10 @@
 // ----------------------------------------------------------------------------------
 
 using Microsoft.Azure.Commands.Common.Authentication.Abstractions;
+#if NETSTANDARD
+using Microsoft.Azure.Commands.Common.Authentication.Abstractions.Core;
```
markcowl: This looks good.  Do we need to change the name of the dll as well?

PR #98, opened 2018-11-27
Test controller update

vladimir-shcherbakov:
Zero guid replaced with dummy guid for the  tenantId to avoid wrong Azure env exception from KeyVault cmdlets. pre and post action params added to runScript method.

### DIFF
--- a/src/TestFx/TestManager.cs
+++ b/src/TestFx/TestManager.cs
@@ -37,7 +37,9 @@ namespace Microsoft.Azure.Commands.TestFx
         private readonly string _callingClassName;
         private string _projectSubfolderForTestsName = null;
         private string _newPsScriptFilename = null;
-        private Dictionary<string, string> _userAgentsToIgnore;
+        private Dictionary<string, string> _matcherExtraUserAgentsToIgnore;
+        private Dictionary<string, string> _matcherNewUserAgentsToIgnore;
+        private Dictionary<string, string> _matcherResourceProviders;
         protected EnvironmentSetupHelper Helper;
         protected readonly List<string> RmModules;
         protected readonly List<string> CommonPsScripts = new List<string>();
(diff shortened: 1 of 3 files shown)


src/TestFx/TestManager.cs
```diff
@@ -273,20 +301,26 @@ protected void SetupAzureContext()
...
-            var resourceProviders = new Dictionary<string, string>
-            {
-                {"Microsoft.Resources", null},
-                {"Microsoft.Features", null},
-                {"Microsoft.Authorization", null},
-                {"Providers.Test", null},
-            };
-
-            var userAgentsToIgnore = new Dictionary<string, string>
+            var resourceProviders = _matcherResourceProviders?.Count > 0
```
MiYanni:
```suggestion
            var resourceProviders = _matcherResourceProviders?.Any()
```

src/TestFx/TestManager.cs
```diff
@@ -273,20 +301,26 @@ protected void SetupAzureContext()
...
+
+            var extraUserAgentsToIgnore = new Dictionary<string, string> // default
             {
                 {"Microsoft.Azure.Management.Resources.ResourceManagementClient", "2016-02-01"},
             };
 
-            _userAgentsToIgnore?.Keys.ForEach(k=> userAgentsToIgnore.Add(k, _userAgentsToIgnore[k]));
+            _matcherExtraUserAgentsToIgnore?.Keys.ForEach(k => extraUserAgentsToIgnore.Add(k, _matcherExtraUserAgentsToIgnore[k])); //extra
+
+            var userAgentsToIgnore = _matcherNewUserAgentsToIgnore?.Count > 0
```
MiYanni:
```suggestion
            var userAgentsToIgnore = _matcherNewUserAgentsToIgnore?.Any()
```

PR #99, opened 2018-11-30
Fix incorrect output of Get-Location when using UseCanonical

number213:
If `$resourceType` passed in `Get-Location` isn't found and `UseCanonical` is used, the function returns null instead of default location.

Since no resource type was found, `locations` array is `$null`. So when each element of `locations` is being normalized through `ForEach-Object` the following statement returns non-empty array: `$null | ForEach-Object`. Because of that, if-statement `$locations.Count -ne 0` is true and `$locations[0]` returns `@($null)[0]` i.e. `$null`

Example:

```
PS C:\> Get-Location "Microsoft.Network" "Usages" "westus"
West US
PS C:\> Get-Location "Microsoft.Network" "Usages" "westus" -UseCanonical

```

### DIFF
--- a/src/ScenarioTest.ResourceManager/Common.ps1
+++ b/src/ScenarioTest.ResourceManager/Common.ps1
@@ -603,7 +603,7 @@ function Get-Location
         }
     }
     $locations = $resourceTypes.Locations
-    if($UseCanonical)
+    if($UseCanonical -and $locations -ne $null)
     {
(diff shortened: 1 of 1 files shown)


MiYanni:
@number213 Good find. I'll be making a PR to this repo this week. I'll merge this and my PR, and cut another build. I'll have to coordinate when we consume the new package since it won't include net452 assemblies.

PR #105, opened 2018-12-07
Moving AzureRMCmdlet and similar classes out of ResourceManager client package

markcowl:


src/ResourceManager/Version2016_09_01/ProviderRegistrationFactory.cs
```diff
@@ -0,0 +1,100 @@
...
+            /// <summary>
+            /// Contains all providers we have attempted to register 
+            /// </summary>
+            private HashSet<string> _registeredProviders;
+
+            private Func<ResourceManagementClient> _createClient;
+
+            private Action<string> _writeDebug;
+
+            private ResourceManagementClient Client => _createClient();
```
MiYanni: Are you intending to create a new client every time the Client property is accessed?

markcowl: It should create a client each time, as the authentication details may change between usages

PR #113, opened 2019-01-02
[Storage] Add support for new XSCL 9.4.2

blueww:
Add a new adapt for support new XSCL.
The change from adapt.8 is: 
 Add  `extern alias` for differenct version of XSCL, and use it in following using section.

### DIFF
--- a/src/Storage/Adapters/WindowsAzure.Storage.9/AzureContextAdapterExtensions.cs
+++ b/src/Storage/Adapters/WindowsAzure.Storage.9/AzureContextAdapterExtensions.cs
@@ -1,0 +1,135 @@
+﻿// ----------------------------------------------------------------------------------
+//
+// Copyright Microsoft Corporation
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// ----------------------------------------------------------------------------------
+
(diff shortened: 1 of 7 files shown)


src/Storage/Adapters/WindowsAzure.Storage.9/AzureContextAdapterExtensions.cs
```diff
@@ -0,0 +1,135 @@
+﻿// ----------------------------------------------------------------------------------
```
markcowl: These changes shouyld not go in the common repo - they shouls dimply go in the storage module in the azure-powershell repo

src/Storage/Adapters/WindowsAzure.Storage.9/AzureContextAdapterExtensions.cs
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        /// <summary>
+        /// Get a storage context client for the storage account represented by this service
+        /// </summary>
+        /// <param name="service">The storage service</param>
+        /// <returns>A CloudStorageAccount client for storage data plane tasks</returns>
+        public static IStorageContext GetStorageContext(this IStorageService service)
+        {
+            return new AzureStorageContext(new CloudStorageAccount(new StorageCredentials(service.Name, service.AuthenticationKeys.First()),
```
markcowl: We need to make sure this fills out the connection string and all endpoints (even if these are filled with dummy values) included in IStorageContext

src/Storage/Adapters/WindowsAzure.Storage.9/AzureStorageContext.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// <summary>
+        /// Get the connection string for this storage account
+        /// </summary>
+        public string ConnectionString {
+            get
+            {
+                string returnValue = null;
+                if (this.StorageAccount != null)
+                {
+                    returnValue = this.StorageAccount.ToString(true);
```
markcowl: WIll this be parseable by earlier versions of storage?

PR #122, opened 2019-01-10
Fix storage script

maddieclayton:


maddieclayton:
This will break azure-powershell if not combined with its appropriate PR, so this should be merged lasted and the PR with the changes related in azure-powershell do the bump

PR #125, opened 2019-01-16
Add Policy client 2018-05-01

MiYanni:
We are adding the 2018-05-01 Policy client and models to the 2018-05-01 resource manager. These are part of the resource manager in the actual SDK for net repo. Currently, GuestConfiguration module needs these models.


MiYanni:
sign: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/ps-common-sign/271/

PR #127, opened 2019-01-17
Add extension method to throw exception when path can't be resolved

cormacpayne:
Used to resolve issue https://github.com/Azure/azure-powershell/issues/2370

Add additional extension method to resolve a path provided, but don't catch the exception like the existing `TryResolvePath` method.


cormacpayne:
Sign job: https://azuresdkci.westus2.cloudapp.azure.com/job/ps-common-sign-esrp/13/

src/Common/Extensions/CmdletExtensions.cs
```diff
@@ -232,6 +232,11 @@ public static string TryResolvePath(this PSCmdlet psCmdlet, string path)
             }
         }
 
+        public static string TryResolvePathWithException(this PSCmdlet psCmdlet, string path)
```
markcowl: Generally, the pattern is that a method with the name 'Try*' does not throw exceptions.  What is this method actually doing?

PR #128, opened 2019-01-17
[Fix] issue #7876: 'Connect-AzAccount -MSI` does not work in the Azure Web App Sandbox

vladimir-shcherbakov:
Short-term fix for the issue https://github.com/Azure/azure-powershell/issues/7876


markcowl:
@vladimir-shcherbakov There are merge conflicts, please fix.

src/Authentication/Authentication/ManagedServiceAccessTokenWebAppsMsiAdHoc.cs
```diff
@@ -0,0 +1,132 @@
...
+{
+    public class ManagedServiceAccessTokenWebAppsMsiAdHoc : IRenewableToken
+    {
+        private readonly IAzureAccount _account;
+        private const string ResourceId = @"https://management.azure.com/";
+        private readonly IHttpOperations<ManagedServiceTokenInfoWebAppsMsiAdHoc> _tokenGetter;
+        private DateTime _expiration = DateTime.UtcNow;
+        private string _accessToken;
+
+        public ManagedServiceAccessTokenWebAppsMsiAdHoc(IAzureAccount account, IAzureEnvironment environment, string tenant = "Common")
```
markcowl: It seems like, if we create a common interface between the MSI payload types, we can reuse the existing ManagedServiceAccessToken - it would be nice to reuse the code - whether that is by adding a new constructor or introducing an abstract base class.

src/Authentication.Abstractions/AzureAccount.cs
```diff
@@ -135,10 +135,16 @@ public static class Property
         /// </summary>
         MSILoginSecret = "MSILoginSecret",
 
+        /// <summary>
+        /// Flag to recognize WebApps ad hoc
+        /// </summary>
+        MSIWebAppsAdHoc = "MSIWebAppsAdHoc",
```
markcowl: us 'AppServiceManagedIdentity' instead of 'MSIWebAppsAdHoc'

src/Authentication/Authentication/ManagedServiceTokenInfoWebAppsMsiAdHoc.cs
```diff
@@ -0,0 +1,53 @@
...
+
+using System;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Commands.Common.Authentication
+{
+    /// <summary>
+    /// Wire representation of MSI token WebApps ad hoc
+    /// </summary>
+    public class ManagedServiceTokenInfoWebAppsMsiAdHoc : ICacheable
```
markcowl: name as above

MiYanni:
As per this PR, these changes need to be made in azure-powershell, and not in this repo. https://github.com/Azure/azure-powershell/pull/8330

MiYanni:
My bad. The Authenication.Abstractions change still needs to be in this repo. All other changes are in azure-powershell.

vladimir-shcherbakov:
I moved the property name out of the project - so don't need the PR.

PR #146, opened 2019-07-15
Add new version of Subscriptions client to ARM common code

cormacpayne:
New version of Subscriptions client is based on the change made in [this spec PR](https://github.com/Azure/azure-rest-api-specs/pull/6473), which adds additional properties to the object returned from the `Tenants_List` operation.


markcowl:
@cormacpayne Do you have a build using this code?  Changes are LGTM

cormacpayne:
@markcowl I have verified with a separate build that the added tenant properties are available in the new API version of the Subscriptions client exposed here

PR #149, opened 2019-08-05
change the deprecated licenseUrl element to license file instead

xiaohuyyyy-ms:
The "PackageLicenseUrl" element in .csproj is deprecated since .net 15.9, it's changed to "PackageLicenseFile" using LINCENSE.txt in the repo instead.

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -16,24 +16,5 @@
     <Exec Command="dotnet msbuild build/pack.proj $(CommandArguments)" />
   </Target>
 
-  <Target Name="CpLicense">
-    <Exec Command="copy LICENSE.txt src\Aks" />
-    <Exec Command="copy LICENSE.txt src\Authentication.Abstractions" />
-    <Exec Command="copy LICENSE.txt src\Authorization" />
-    <Exec Command="copy LICENSE.txt src\Common" />
-    <Exec Command="copy LICENSE.txt src\Compute" />
-    <Exec Command="copy LICENSE.txt src\Graph.Rbac" />
-    <Exec Command="copy LICENSE.txt src\KeyVault" />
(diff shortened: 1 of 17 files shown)


build.proj
```diff
@@ -16,5 +16,24 @@
     <Exec Command="dotnet msbuild build/pack.proj $(CommandArguments)" />
   </Target>
 
+  <Target Name="CpLicense">
```
erich-wang: remove this target, use link file in csproj instead

src/Aks/Aks.csproj
```diff
@@ -49,4 +49,8 @@
     </EmbeddedResource>
   </ItemGroup>
 
+  <ItemGroup>
+    <None Include="LICENSE.txt" Pack="true" Visible="false" PackagePath=""/>
```
erich-wang:
```
<None Include="..\..\LICENSE.txt" Link="LICENSE.txt" Pack="true" PackagePath="" />
```

xiaohuyyyy-ms:
This commit is for dotnet version15.9 and above. As for version 15.8, this code does not work, since the jenkins job is currently using dotnet version 15.8, this commit will result in the failure of the signing job in jenkins ("ps-common-sign“).

markcowl:
Should need a simultaneous update of the packaging job - perhaps we cna couple this with the packing job being in ADO.  I assume this has to do with NuGet and not .Net particularly (although dotnet pack wil use a particular version of nuget).

markcowl:
@erich-wang  Need to verify a signed build with this change this against PowerShell to verify

erich-wang:
Close it as it is not unnecessary now.

PR #150, opened 2019-08-14
Update changelog (testing)

markcowl:


markcowl:
@isra-fel Looks like one of the ADO checks is not running - can you verify this is the intended result?

isra-fel:
> @isra-fel Looks like one of the ADO checks is not running - can you verify this is the intended result?

I believe it's the "Release+Pack" job you mean.
The reason that job wasn't run is it requires the `Configuration` variable to equal "Release" or "All" but the default value is "Debug".
Appearently this pipeline is trying to do two things -- build+test vs. build+pack -- in a "switch-case" manner, but we already have pipelines for the latter (ps-common-sign, ps-common-build). So I remove the unrun job.

BTW I saw there was a devdiv.azure.com pipeline in the checks. And it was doing the exact same thing as the dev.azure.com pipeline. So I paused it and eventually we can remove it.

markcowl:
@isra-fel  will open a new test PR then, and hopefully all the checks line up

PR #151, opened 2019-08-27
[WIP] Common functions for SKU completer

isra-fel:
## What I did

Some common (helper) functions that are shared by the SKU completers for:
- hashing context
- building argument completer script
- reading paged data (`IPage<>` interface)

## How to use
See the SKU completers in https://github.com/Azure/azure-powershell/pull/9818


src/ResourceManager/Version2016_09_01/ArgumentCompleters/ArgumentCompleterHelper.cs
```diff
@@ -0,0 +1,106 @@
...
+
+    /// <summary>
+    /// Provides helper functions and classes that other argument completers can use.
+    /// </summary>
+    public static class ArgumentCompleterHelper
+    {
+        /// <summary>
+        /// Builds the actual ps script for argument completer.
+        /// </summary>
+        public class ScriptBuilder
```
markcowl: Why is this non-static class encapsulated in a static class?  Why not ximply have a ScriptBuilder class.  Normally classes are only encapsulated to implement the AbstractFactory pattern, singleton pattern, , or similar patterns where the detail of the underlying structure is necessary

PR #163, opened 2020-01-08
Add FileExists property to IFileProvider

rlrossiter:
This adds a FileExists property to IFileProvider to bring it closer parity with `System.IO.File`


rlrossiter:
A bit of background on this change:

I'm looking to add SSH-like functionality to azure-powershell. In order to do this, I will be provided 2 file paths for RSA SSH keys (public/private key pair). I need to check if the files exist before proceeding, and I'd like to be able to unit test it.

Things I'm not sure about:
  - How do we version this change on the package?
    - it's an additive change, but breaks compilation does that require a major version bump?
  - `IFileProvider` vs. `IFileSystem`
    - I added this to IFileProvider because I see that's in use by azure-powershell, not sure if this should be part of IFileSystem as well

dingmeng-xue:
Close this PR after confirmation

PR #164, opened 2020-01-10
Add SSH authentication abstractions

rlrossiter:
New authentication abstractions are needed in order to acquire SSH certificates. These are represented in the same way as Access Tokens (`IAccessToken`), but now the AccessToken property is populated with SSH certificate data. Because of this, the authentication functions are similar in signature, with only a new parameters interface added. These parameters contain the SSH key data that need to be associated with the authenticating user.

This new authentication factory interface was also added to the `AzureSession`.


rlrossiter:
FYI, The implementation of the `ISSHAuthenticationFactory` is heavily dependent on MSAL, so these interfaces will not have a powershell implementation until powershell is on MSAL.

dingmeng-xue:
Close this PR after confirmation

PR #166, opened 2020-01-13
Remove Test in Release mode

msJinLei:
Test build error in release mode so that remove them.


isra-fel:
sign job: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=231842

PR #168, opened 2020-02-08
Add az version, time, and subscription id to client telemetry

dingmeng-xue:
Changes around customer data in client telemetry:

1. Fetch and add az version to user agent
2. Add start time, end time, and effective execution duration
3. Add AZUREPS_HOST_ENVIRONMENT to user agent
4. Add subscription id and tenant id

A minor change is to remove exception message if cmdlet has no exception.
Another minor change is fixing issue https://github.com/Azure/azure-powershell/issues/10887

### DIFF
--- a/src/Common/AzurePSCmdlet.cs
+++ b/src/Common/AzurePSCmdlet.cs
@@ -27,6 +27,8 @@ using System.Linq;
 using System.Management.Automation;
 using System.Text;
 using System.Collections.Generic;
+using System.Management.Automation.Runspaces;
+using System.Collections.ObjectModel;
 
 namespace Microsoft.WindowsAzure.Commands.Utilities.Common
 {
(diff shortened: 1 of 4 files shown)


src/Common/AzurePowerShell.cs
```diff
@@ -22,7 +22,7 @@ public class AzurePowerShell
     {
         public const string AssemblyCompany = "Microsoft";
 
-        public const string AssemblyProduct = "Microsoft Azure Powershell";
+        public const string AssemblyProduct = "Microsoft Azure PowerShell";
```
isra-fel: Did we talk about keeping the wrong spelling as-is, because we don't want the telemetry data to have both spelling?

src/Common/MetricHelper.cs
```diff
@@ -24,6 +24,7 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Management.Automation.Host;
+using System.Net.Http.Headers;
```
isra-fel: Is This namespace not used?

src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
```diff
@@ -334,13 +346,6 @@ protected override void LogCmdletStartInvocationInfo()
             }
         }
 
-        protected override void LogCmdletEndInvocationInfo()
```
isra-fel: cool thanks

src/Common/AzurePSCmdlet.cs
```diff
@@ -27,6 +27,8 @@
 using System.Management.Automation;
 using System.Text;
 using System.Collections.Generic;
+using System.Management.Automation.Runspaces;
+using System.Collections.ObjectModel;
```
erich-wang: suggest to run "Remove and Sort Usings" Ctrl+R, Ctrl+G in Visual Studio

src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
```diff
@@ -295,6 +296,17 @@ protected override void InitializeQosEvent()
                 ParameterSetName = this.ParameterSetName
             };
 
+            if (AzVersion == null)
+            {
+                AzVersion = this.LoadAzVersion();
+                UserAgent = new ProductInfoHeaderValue("AzurePowershell", string.Format("Az{0}", AzVersion)).ToString();
+                string HostEnv = Environment.GetEnvironmentVariable("AZUREPS_HOST_ENVIRONMENT");
```
erich-wang: suggest local variable starting with lower case, i.e. hostEnv

src/Common/AzurePowerShell.cs
dingmeng-xue: It's a different one. Product name is used for branding.

PR #178, opened 2020-03-06
Update default context check option

dingmeng-xue:


src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
```diff
@@ -172,7 +173,7 @@ protected override string DataCollectionWarning
         /// Whether this cmdlet requires default context.
         /// If false, the logic of referencing default context would be omitted.
         /// </summary>
-        protected virtual bool RequireDefaultContext => true;
+        protected virtual bool RequireDefaultContext() { return true; }
```
isra-fel: We've already been using this property in two cmdlets. Changing it to a method will require us to change those cmdlets as well. Would you want me to do that?

isra-fel: For example [here](https://github.com/Azure/azure-powershell/blob/msal-build/src/Accounts/Accounts/Account/ConnectAzureRmAccount.cs#L185) we need to do the following update
```diff
-        protected override bool RequireDefaultContext => false;
+        protected override bool RequireDefaultContext() { return false; }
```

dingmeng-xue: yes. your original approach doesn't work. You can test too. No idea the syntax you are using. but my approach is the safe way and I verified it. I haven't submitted my commits in azure-powershell.

dingmeng-xue: I write a simple example. It shows your code is correct. I'd like to change them back to follow convention if there is no other problem.

PR #179, opened 2020-03-10
AzureRestClient support

dingmeng-xue:


src/Common/AzureRest/AzureRestOperations.cs
```diff
@@ -0,0 +1,295 @@
...
+            _httpRequest.Method = method;
+            _httpRequest.RequestUri = new System.Uri(_url);
+            // Set Headers
+            if (Client.GenerateClientRequestId != null && Client.GenerateClientRequestId.Value)
+            {
+                _httpRequest.Headers.TryAddWithoutValidation("x-ms-client-request-id", System.Guid.NewGuid().ToString());
+            }
+            if (Client.AcceptLanguage != null)
+            {
+                if (_httpRequest.Headers.Contains("accept-language"))
```
erich-wang: suggest to define constant string to avoid duplication

src/Common/AzureRest/AzureRestOperations.cs
```diff
@@ -0,0 +1,295 @@
...
+                }
+                if (_shouldTrace)
+                {
+                    ServiceClientTracing.Error(_invocationId, ex);
+                }
+                _httpRequest.Dispose();
+                if (_httpResponse != null)
+                {
+                    _httpResponse.Dispose();
+                }
```
erich-wang: use httpResponse?.Dispose() for less code :)

src/Common/AzureRest/AzureRestOperations.cs
```diff
@@ -0,0 +1,295 @@
...
+            _result.Request = _httpRequest;
+            _result.Response = _httpResponse;
+            if (_httpResponse.Headers.Contains("x-ms-request-id"))
+            {
+                _result.RequestId = _httpResponse.Headers.GetValues("x-ms-request-id").FirstOrDefault();
+            }
+            // Deserialize Response
+            if ((int)_statusCode == 200)
+            {
+                _responseContent = await _httpResponse.Content.ReadAsStringAsync().ConfigureAwait(false);
```
erich-wang: Is it possible responseConent is null? If yes, we should skip DeserializeObject below if it is null

src/Common/AzureRest/IAzureRestOperations.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// <param name="content">The content in request body</param>
+        /// <param name="customHeaders">The headers that will be added to request.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <exception cref="Microsoft.Rest.Azure.CloudException">
+        /// Thrown when the operation returned an invalid status code
+        /// </exception>
+        /// <exception cref="Microsoft.Rest.ValidationException">
+        /// Thrown when a required parameter is null
+        /// </exception>
+        Task<AzureOperationResponse<T>> BeginHttpMessagesAsync<T>(HttpMethod method, string path, Dictionary<string, List<string>> queries = null, string fragment = null, string content = null, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default);
```
erich-wang: recommend to use interface instead of class for parameters: e.g. IDictionary<>, IList<>

src/Common/AzureRest/AzureRestOperations.cs
```diff
@@ -0,0 +1,295 @@
...
+            if (path == null)
+            {
+                throw new ValidationException(ValidationRules.CannotBeNull, "path");
+            }
+            if (method == null)
+            {
+                throw new ValidationException(ValidationRules.CannotBeNull, "method");
+            }
+            // Tracing
+            bool _shouldTrace = ServiceClientTracing.IsEnabled;
```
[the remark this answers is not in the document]
dingmeng-xue: Consist with auto generated code. If there is any issue, we can easily compare with the latest generated code.

src/Common/AzureRest/AzureRestOperations.cs
```diff
@@ -0,0 +1,295 @@
...
+            if (_shouldTrace)
+            {
+                _invocationId = ServiceClientTracing.NextInvocationId.ToString();
+                Dictionary<string, object> tracingParameters = new Dictionary<string, object>();
+                tracingParameters.Add("cancellationToken", cancellationToken);
+                ServiceClientTracing.Enter(_invocationId, this, "BeginDelete", tracingParameters);
+            }
+            // Construct URL
+            var _baseUrl = Client.BaseUri.AbsoluteUri;
+            var _url = new System.Uri(_baseUrl + path).ToString();
```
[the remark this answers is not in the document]
dingmeng-xue: It's a convention used by SDK. If we need to check that, we can do that on powershell layer.

src/Common/AzureRest/AzureRestOperations.cs
dingmeng-xue: copied from SDK generated code.

src/Common/AzureRest/AzureRestOperations.cs
dingmeng-xue: copied from SDK generated code.

src/Common/AzureRest/AzureRestOperations.cs
dingmeng-xue: HTTP 200 should always return content. I can add a check

dingmeng-xue:
@erich-wang , I have revised code according to comments, except codes which is copied from generated SDK. I keep the consistency and will be helpful to update accordingly in the future.

PR #180, opened 2020-03-19
Fix issues in AzureRestClient

dingmeng-xue:
* Change content from string to object
* Add deserialize settings
* Add Page support

### DIFF
--- a/src/Common/AzureRest/AzureRestOperations.cs
+++ b/src/Common/AzureRest/AzureRestOperations.cs
@@ -38,12 +38,24 @@ namespace Microsoft.WindowsAzure.Commands.Common.AzureRest
         /// </summary>
         public AzureRestClient Client { get; private set; }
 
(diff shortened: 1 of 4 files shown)


src/Common/AzureRest/AzureRestOperations.cs
```diff
@@ -276,26 +297,26 @@ public List<T> GetResouceList<T>(string resourceUri, string apiVersion, IDiction
             return BeginHttpGetMessagesAsync<List<T>>(resourceUri, apiVersion, customHeaders, cancellationToken).GetAwaiter().GetResult();
         }
 
-        public IPage<T> GetResoucePage<T>(string resourceUri, string apiVersion, IDictionary<string, IList<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken))
+        public P GetResoucePage<P, T>(string resourceUri, string apiVersion, IDictionary<string, IList<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken)) where P : IPage<T>
```
erich-wang: could we just use IPage<T> GetResourcePage<T>?

dingmeng-xue: As discussed, I need to pass implementation into this method for JSON deserialization

PR #181, opened 2020-03-19
Add attribute for preview cmdlet

dingmeng-xue:


src/Common/CustomAttributes/PreviewAttributeHelper.cs
```diff
@@ -0,0 +1,87 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Management.Automation;
+using System.Reflection;
+
+namespace Microsoft.WindowsAzure.Commands.Common.CustomAttributes
+{
+    public class PreviewAttributeHelper
```
erich-wang: Suggest to use internal if this class won't be used by other library.

src/Common/CustomAttributes/CmdletPreviewAttribute.cs
```diff
@@ -0,0 +1,48 @@
...
+using Microsoft.WindowsAzure.Commands.Common.Properties;
+using System;
+using System.Management.Automation;
+
+namespace Microsoft.WindowsAzure.Commands.Common.CustomAttributes
+{
+    [AttributeUsage(
+     AttributeTargets.Class,
+     AllowMultiple = true)]
+    public class CmdletPreviewAttribute : System.Attribute
```
erich-wang: I guess this attribute will be declared at most once for each class, please use **AllowMultiple = false** if this is the case.

PR #192, opened 2020-07-15
Fix ResourceIdentifier parsing issue for subresource's name

msJinLei:
issue: https://github.com/Azure/azure-powershell/issues/11692

### DIFF
--- a/src/ResourceManager/Version2016_09_01/Utilities/Models/ResourceIdentifier.cs
+++ b/src/ResourceManager/Version2016_09_01/Utilities/Models/ResourceIdentifier.cs
@@ -14,6 +14,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 
 namespace Microsoft.Azure.Management.Internal.Resources.Utilities.Models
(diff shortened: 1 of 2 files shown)


src/ResourceManager/Version2016_09_01/Utilities/Models/ResourceIdentifier.cs
```diff
@@ -32,42 +33,59 @@ public class ResourceIdentifier
...
+                providers = "providers";
+        }
+
         public ResourceIdentifier(string idFromServer)
         {
             if (!string.IsNullOrEmpty(idFromServer))
             {
                 string[] tokens = idFromServer.Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
-                if (tokens.Length < 8)
+                if (tokens.Length % 2 > 0 || tokens.Length < 8)
```
dingmeng-xue, line 48: This assumption may not correct. below are 2 examples
/subscriptions/{}/resourcegroup/{}/providers/{}/resourceType/{resource}/{operation}
/subscriptions/{}/resourcegroup/{}/providers/{}/resourceType/{resource}/subresources

src/ResourceManager/Version2016_09_01/Utilities/Models/ResourceIdentifier.cs
```diff
@@ -32,42 +33,59 @@ public class ResourceIdentifier
 
         public ResourceIdentifier() { }
 
+        private class UrlTagName
+        {
+            public const string subscriptions = "subscriptions",
```
dingmeng-xue, line 38: I don't think we need to add this assumption. If we need to add it, please compare string in case insensitive

PR #200, opened 2020-08-04
Expose property subscriptionPolices in AzureSubscription

BethanyZhou:


src/ResourceManager/Version2016_09_01/Utilities/SubscriptionExtensions.cs
```diff
@@ -29,6 +30,7 @@ public static AzureSubscription ToAzureSubscription(this Subscription other, IAz
             subscription.SetEnvironment(environment != null ? environment.Name : EnvironmentName.AzureCloud);
             subscription.SetHomeTenant(other.TenantId ?? retrievedByTenant);
             subscription.SetTenant(retrievedByTenant);
+            subscription.SetSubscriptionPolicies(JsonConvert.SerializeObject(other.SubscriptionPolicies));
```
isra-fel, line 33: Why are you setting the serialized object instead the raw object into subscription?

BethanyZhou: Because I stored subcriptionPolices in extendedProperty, whose type is IDictionary<string,string>

isra-fel: Makes sense

PR #203, opened 2020-08-11
Merge built-in environments and discovered environemnts

dingmeng-xue:
Fix for this issue

https://github.com/Azure/azure-powershell/issues/12633

### DIFF
--- a/src/Authentication.Abstractions.Test/AzureEnvironmentTests.cs
+++ b/src/Authentication.Abstractions.Test/AzureEnvironmentTests.cs
@@ -59,8 +59,24 @@ namespace Authentication.Abstractions.Test
             var armEnvironments = AzureEnvironment.InitializeBuiltInEnvironments(null, httpOperations: TestOperationsFactory.Create().GetHttpOperations());
 
             // Check AzureCloud is added to public environment list even discovery endpoint doesn't return AzureCloud.
-            Assert.True(AzureEnvironment.PublicEnvironments.ContainsKey(EnvironmentName.AzureCloud));
-            Assert.Equal(AzureEnvironment.TypeBuiltIn, AzureEnvironment.PublicEnvironments[EnvironmentName.AzureCloud].Type);
+            Assert.Equal(4, armEnvironments.Count);
+            Assert.Equal(AzureEnvironment.TypeBuiltIn, armEnvironments[EnvironmentName.AzureCloud].Type);
+            Assert.Equal(AzureEnvironment.TypeBuiltIn, armEnvironments[EnvironmentName.AzureChinaCloud].Type);
+            Assert.Equal(AzureEnvironment.TypeDiscovered, armEnvironments[EnvironmentName.AzureGermanCloud].Type);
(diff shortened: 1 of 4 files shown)


src/Authentication.Abstractions.Test/TestData/ArmResponseNoAzureCloud.json
```diff
@@ -0,0 +1,57 @@
+[
+  {
+    "portal": "https://portal.azure.us",
```
markcowl, line 3: I think we might want to add a specific new piece of data with only one enbtry, just for completeness.  We want to make sure we protect ourselves from future bugs where we make assumptions about this payload.

src/Authentication.Abstractions.Test/AzureEnvironmentTests.cs
```diff
@@ -52,6 +52,17 @@ public void TestArmCloudMetadata20190501Init()
...
         }
 
+        [Fact]
+        public void TestArmResponseNoAzureCloud()
+        {
+            Environment.SetEnvironmentVariable(ArmMetadataEnvVariable, @"TestData\ArmResponseNoAzureCloud.json");
+            var armEnvironments = AzureEnvironment.InitializeBuiltInEnvironments(null, httpOperations: TestOperationsFactory.Create().GetHttpOperations());
+
+            // Check AzureCloud is added to public environment list even discovery endpoint doesn't return AzureCloud.
+            Assert.True(AzureEnvironment.PublicEnvironments.ContainsKey(EnvironmentName.AzureCloud));
```
markcowl, line 62: I would assert this for all of the built-in environments

src/Authentication.Abstractions.Test/TestData/ArmResponseNoAzureCloud.json
```diff
@@ -0,0 +1,57 @@
+[
+  {
```
markcowl, line 2: I also got a report on the same thread that there are additional errors that prevent module load when customers use the  ```ARM_CLOUD_METADATA_URL``` environment variable to point at a metadata endpoint that does not have the 'AzureCloud' entry.  I believe we have a test using the environment variable for discovery, and we should ensure that it also convers this scenario.

dingmeng-xue: You are right. The discovery approach via default DNS, parameter, or environment variable are going into the same logic. It should be covered.

src/Authentication.Abstractions.Test/AzureEnvironmentTests.cs
dingmeng-xue: Sure. I will add them.

src/Authentication.Abstractions.Test/TestData/ArmResponseNoAzureCloud.json
dingmeng-xue: Sure. I will add it.

dingmeng-xue:
AzureEnvironment.InitializeBuiltInEnvironments is an internal method and public cloud map will be populated after this method. So, in the test case, check public environment variable is not correct because only InitializeBuiltInEnvironments is called. I only check public cloud environment is correctly returned in test cases.

PR #218, opened 2020-09-15
Add support for Azure.Identity integration

erich-wang:
Add support for Azure.Identity integration


src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
```diff
@@ -489,5 +505,46 @@ public bool ShouldGetByName(string resourceGroupName, string name)
 
             return false;
         }
+
+        private event EventHandler<StreamEventArgs> _writeDebugEvent;
```
erich-wang, line 509: double check if this is needed?

isra-fel (approved):
LGTM

PR #219, opened 2020-09-25
fix DateTime parse issue

BethanyZhou:


src/Common/CustomAttributes/GenericBreakingChangeAttribute.cs
```diff
@@ -68,8 +68,11 @@ public GenericBreakingChangeAttribute(string message, string deprecateByVersion,
             this.DeprecateByVersion = deprecateByVersion;
             this.DeprecateByVersionSet = true;
 
-            this.ChangeInEfectByDate = DateTime.Parse(changeInEfectByDate);
-            this.ChangeInEfectByDateSet = true;
+            if (DateTime.TryParse(changeInEfectByDate, out DateTime result))
```
erich-wang, line 71:
```suggestion
            if (DateTime.TryParse(changeInEfectByDate, new CultureInfo("en-US"), DateTimeStyles.None, out DateTime result))
```

BethanyZhou:
@erich-wang Updated according to your suggestion, thanks

src/Common/CustomAttributes/GenericBreakingChangeAttribute.cs
```diff
@@ -68,8 +69,12 @@ public GenericBreakingChangeAttribute(string message, string deprecateByVersion,
             this.DeprecateByVersion = deprecateByVersion;
             this.DeprecateByVersionSet = true;
 
-            this.ChangeInEfectByDate = DateTime.Parse(changeInEfectByDate);
-            this.ChangeInEfectByDateSet = true;
+            string[] formats = { "MM/dd/yyyy", "MM/d/yyyy", "M/dd/yyyy", "M/d/yyyy" };
```
erich-wang, line 72: does it work with "9/27/20"?

BethanyZhou: No...Should we support this format?

erich-wang (approved):

PR #221, opened 2020-10-09
Add ACR domain suffix

VeryEarly:


src/Authentication.Abstractions/AzureEnvironment.cs
```diff
@@ -100,6 +100,7 @@ public class AzureEnvironment : IAzureEnvironment, IEquatable<AzureEnvironment>
                 GraphEndpointResourceId = AzureEnvironmentConstants.AzureGraphEndpoint,
                 DataLakeEndpointResourceId = AzureEnvironmentConstants.AzureDataLakeServiceEndpointResourceId,
                 BatchEndpointResourceId = AzureEnvironmentConstants.BatchEndpointResourceId,
+                AzureContainerRegistryEndpointSuffix = AzureEnvironmentConstants.AzureContainerRegistryEndpointSuffix,
```
dingmeng-xue, line 103: Azure is not required. Just ContainerResigtryEndpointSuffix.

src/Common/RecordingTracingInterceptor.cs
```diff
@@ -19,18 +19,23 @@
...
 using System.Net.Http;
+using System.Text.RegularExpressions;
 
 namespace Microsoft.Azure.ServiceManagement.Common.Models
 {
     public class RecordingTracingInterceptor : Hyak.Common.ICloudTracingInterceptor
     {
-        public RecordingTracingInterceptor(ConcurrentQueue<string> queue)
+
+        public RecordingTracingInterceptor(ConcurrentQueue<string> queue, IList<Regex> matchers = null)
```
dingmeng-xue, line 29: Keep original constructor to ensure capability

src/Common/Utilities/GeneralUtilities.cs
```diff
@@ -252,23 +252,27 @@ public static string EnsureTrailingSlash(string url)
             return address.Uri.AbsoluteUri;
         }
 
-        public static string GetHttpResponseLog(string statusCode, IDictionary<string, IEnumerable<string>> headers, string body)
+        public static string GetHttpResponseLog(string statusCode, IDictionary<string, IEnumerable<string>> headers, string body, IList<Regex> matchers = null)
```
dingmeng-xue, line 255: Keep original method and add new method with new input parameter

src/Common/RecordingTracingInterceptor.cs
VeryEarly: this will be compatible with the original constructor

src/Common/Utilities/GeneralUtilities.cs
VeryEarly: this will be compatible with the original

src/Common/AzurePSCmdlet.cs
```diff
@@ -46,6 +48,9 @@ public abstract class AzurePSCmdlet : PSCmdlet, IDisposable
         private object lockObject = new object();
         private AzurePSDataCollectionProfile _cachedProfile = null;
 
+        private IList<Regex> _matchers;
+        private Regex _defaultMatcher = new Regex("(\\s*\"refresh_token\"\\s*:\\s*)\"[^\"]+\"");
```
dingmeng-xue, line 52: Shall it be static?

dingmeng-xue (dismissed):

dingmeng-xue (approved):

PR #222, opened 2020-10-12
Set X509KeyStorageFlags.PersistKeySet flag explicitly

dingmeng-xue:
Dotnet had a bug that did not remove temporary private keys (non-persistent) for certificates. September Rollup fixed this bug. As the result, client is blocked because private keys are removed during importing publish setting because its code uses default key setting. The correct way is we should set PersistKeySet explicitly

Details is from https://docs.microsoft.com/en-us/troubleshoot/dotnet/framework/install-pfx-file-using-x509certificate

### DIFF
--- a/src/Authentication.Abstractions/Properties/AssemblyInfo.cs
+++ b/src/Authentication.Abstractions/Properties/AssemblyInfo.cs
@@ -43,5 +43,5 @@ using System.Runtime.InteropServices;
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("1.0.0.1")]
+[assembly: AssemblyVersion("1.0.0.0")]
 [assembly: AssemblyFileVersion("1.0.0.1")]
--- a/src/ServiceManagement/PublishSettingsImporter.cs
+++ b/src/ServiceManagement/PublishSettingsImporter.cs
@@ -91,7 +91,7 @@ namespace Microsoft.Azure.ServiceManagemenet.Common
                 certificateString = profile.ManagementCertificate;
             }
 
-            X509Certificate2 certificate = new X509Certificate2(Convert.FromBase64String(certificateString), string.Empty);
+            X509Certificate2 certificate = new X509Certificate2(Convert.FromBase64String(certificateString), string.Empty, X509KeyStorageFlags.PersistKeySet);
             AzureSession.Instance.DataStore.AddCertificate(certificate);
 
             return certificate;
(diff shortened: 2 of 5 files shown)


src/ServiceManagement/PublishSettingsImporter.cs
```diff
@@ -91,7 +91,7 @@ private static PublishData DeserializePublishData(Stream stream)
                 certificateString = profile.ManagementCertificate;
             }
 
-            X509Certificate2 certificate = new X509Certificate2(Convert.FromBase64String(certificateString), string.Empty);
+            X509Certificate2 certificate = new X509Certificate2(Convert.FromBase64String(certificateString), string.Empty, X509KeyStorageFlags.PersistKeySet);
```
markcowl, line 94: I think we can do this in a more targeted way by making the change here:https://github.com/Azure/azure-powershell-common/blob/master/src/Authentication.Abstractions/DiskDataStore.cs#L223

This would involve only one assembly, which is guaranteed to be backwards compatible, where this one is not.

The idea would be to make a change like this before the certificate is stored:

``` c#
X509Certificate2 storedCertificate = new X509Certificate2(certificate.RawData, null, X509KeyStorageFlags.PersistKeySet);
```

Then store this certificate created from the other.  This would require a test to make sure it resolved the problem, however, the reason to favor this change is because this library is *guaranteed* to be backwards compatible, and only one dll would need to change.

markcowl (changes requested):

src/Authentication.Abstractions/Properties/AssemblyInfo.cs
```diff
@@ -43,5 +43,5 @@
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("1.0.0.1")]
```
markcowl, line 46: This sill likely cause a problem the next time we release this library for Az - we will likely need to put this back to avoid any backwards compatibility issues.

markcowl (approved):

src/Authentication.Abstractions/Properties/AssemblyInfo.cs
dingmeng-xue: The latest common AssemblyVersion is 1.0.0.0. No idea why it was changed to 1.0.0.1 after Azure was released.

dingmeng-xue:
It's not easy to change Microsoft.Azure.Commands.Common.Authentication.Abstractions because this file is used by AzureRM.profile too. AzureRM.Profile\5.8.3\Microsoft.Azure.Commands.Common.Authentication.Abstractions.dll is loaded before file from Azure.

markcowl (approved):

PR #230, opened 2020-11-03
Generate storage management sdk of api version 2019-06-01

msJinLei:
https://github.com/Azure/azure-powershell/issues/13372


blueww (dismissed):
The code looks good for me.
But since I have not generate it before, suggest to ask Mark to take a look.

msJinLei:
> The code looks good for me.
> But since I have not generate it before, suggest to ask Mark to take a look.

Thanks for reviewing!

blueww:
@msJinLei 

I get a question from sql team for create strorage container in their PSH module test.
I think we might should also include create/remove blob container and create/remove file share API in this SDK.
How do you think about it?

isra-fel (approved):

PR #235, opened 2020-12-23
Show correct client request id on debug message

dingmeng-xue:
https://github.com/Azure/azure-powershell/issues/13745

The value of x-ms-client-request-id showing on debug log is different from the real value on request. The root cause is due the the sequence of request id generation, request updating in sdk, and request updating by Azure PowerShell. The order is below:

1. client request id is generated (ID1) in the constructor of cmdlet
2. ServiceClientTracingInterceptor is registered into service client
3. CmdletInfoHandler is registered into http pipeline of service client. ID1 is passed into CmdletInfoHandler.
2. Service client generates request id (ID2) because request doesn't contain x-ms-client-request-id
3. Service client calls ServiceClientTracingInterceptor to print out http request
4. Service client calls sendAsync() and start HTTP pipeline
5. CmdletInfoHandler sets ID1 into request

So, ID2 is printed out on console as debug message but ID1 is the real value used by http request.

Solution is to set ID1 into ServiceClientTracingInterceptor . ID1 is set into http request before being printed out.

Additionally, couple changes are included together
1. RecordingTracingInterceptor is used by data plane cmdlet. Request id is set too.
2. Add extensions to HttpRequestMessage
3. Add null reference check in GetLog() of GeneralUtilities

### DIFF
--- a/src/Common/CmdletInfoHandler.cs
+++ b/src/Common/CmdletInfoHandler.cs
@@ -12,6 +12,7 @@
 // limitations under the License.
 // ----------------------------------------------------------------------------------
 
+using Microsoft.WindowsAzure.Commands.Common.Extensions;
 using System;
 using System.Net.Http;
 using System.Threading;
@@ -64,11 +65,7 @@ namespace Microsoft.WindowsAzure.Commands.Common
             }
             if (ClientRequestId != null)
             {
-                if (request.Headers.Contains(ApiConstants.HeaderNameClientRequestId))
-                {
-                    request.Headers.Remove(ApiConstants.HeaderNameClientRequestId);
-                }
-                request.Headers.TryAddWithoutValidation(ApiConstants.HeaderNameClientRequestId, ClientRequestId);
+                request.AddClientRequestId(ClientRequestId);
             }
             return base.SendAsync(request, cancellationToken);
         }
--- a/src/Common/CmdletInfoHandler.cs
+++ b/src/Common/CmdletInfoHandler.cs
@@ -56,21 +56,26 @@ namespace Microsoft.WindowsAzure.Commands.Common
         {
             if (Cmdlet != null)
             {
-                request.Headers.Add("CommandName", Cmdlet);
+                UpdateRequestHeader(request, "CommandName", Cmdlet);
             }
             if (ParameterSet != null)
             {
-                request.Headers.Add("ParameterSetName", ParameterSet);
+                UpdateRequestHeader(request, "ParameterSetName", ParameterSet);
             }
             if (ClientRequestId != null)
             {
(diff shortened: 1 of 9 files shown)


src/ResourceManager/Version2016_09_01/ServiceClientTracingInterceptor.cs
```diff
@@ -62,8 +65,17 @@ public void ReceiveResponse(string invocationId, System.Net.Http.HttpResponseMes
...
+            // Update request here to ensure its value consistent with real result.
+            if (request != null && clientRequestId != null)
+            {
+                if (request.Headers.Contains(ApiConstants.HeaderNameClientRequestId))
+                {
+                    request.Headers.Remove(ApiConstants.HeaderNameClientRequestId);
+                }
+                request.Headers.TryAddWithoutValidation(ApiConstants.HeaderNameClientRequestId, clientRequestId);
+            }
+            MessageQueue.CheckAndEnqueue(GeneralUtilities.GetLog(request, Matchers));
```
erich-wang, line 78: request might be null here.

src/Common/CmdletInfoHandler.cs
```diff
@@ -64,11 +64,11 @@ protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage reques
...
             {
-                if (request.Headers.Contains("x-ms-client-request-id"))
+                if (request.Headers.Contains(ApiConstants.HeaderNameClientRequestId))
                 {
-                    request.Headers.Remove("x-ms-client-request-id");
+                    request.Headers.Remove(ApiConstants.HeaderNameClientRequestId);
                 }
-                request.Headers.TryAddWithoutValidation("x-ms-client-request-id", ClientRequestId);
+                request.Headers.TryAddWithoutValidation(ApiConstants.HeaderNameClientRequestId, ClientRequestId);
             }
```
erich-wang, line 72: nit: It seems these code are similar in three places, how about move to helper method?

dingmeng-xue: It's current design problem. Ideally, this logic should only show in one place. but that needs cmdlet info handler and trancing inceptors to reference each other.  So, just copy code 3 times.

I don't hope to introduce additional helper class for this problem.

src/ResourceManager/Version2016_09_01/ServiceClientTracingInterceptor.cs
dingmeng-xue: I added check in GetLog method.

src/Common/CmdletInfoHandler.cs
dingmeng-xue: Add extensions

erich-wang (approved):

PR #238, opened 2021-01-07
add Base64UrlHelper

VeryEarly:


src/Common/Utilities/Base64UrlHelper.cs
```diff
@@ -0,0 +1,96 @@
...
+// ----------------------------------------------------------------------------------
+
+using System;
+using System.Globalization;
+using System.Text;
+
+namespace Microsoft.Azure.Commands.Profile.Utilities
+{
+    //The source code is copied from https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/client/Microsoft.Identity.Client/Utils/Base64UrlHelpers.cs
+    internal static class Base64UrlHelper
```
isra-fel, line 22: You probably want to make it public because the consumer will be in a different dll

VeryEarly: sure, forget to change it

isra-fel (approved):

PR #247, opened 2021-02-01
Display warning message that AzureRM will retire

erich-wang:
Two changes in this RP:
1. Display warning message that AzureRM will retire
2. Add missing pieces for  official 1.1.8 version

### DIFF
--- a/src/Common/AzurePSCmdlet.cs
+++ b/src/Common/AzurePSCmdlet.cs
@@ -27,6 +27,7 @@ using System.Linq;
 using System.Management.Automation;
 using System.Text;
 using System.Collections.Generic;
+using Microsoft.WindowsAzure.Commands.Common.Properties;
 
 namespace Microsoft.WindowsAzure.Commands.Utilities.Common
 {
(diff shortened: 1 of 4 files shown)


src/Common/AzurePSCmdlet.cs
```diff
@@ -365,6 +366,32 @@ protected override void BeginProcessing()
...
+            bool supressRetireWarningMessage = false;
+
+            try
+            {
+                supressRetireWarningMessage = bool.Parse(Environment.GetEnvironmentVariable("SuppressAzureRmModulesRetiringWarning"));
+            }
+            catch (Exception)
+            {
+                //no action
+            }
```
BethanyZhou, line 387: How about using tryParse()
bool.TryParse(Environment.GetEnvironmentVariable("SuppressAzureRmModulesRetiringWarning"), out supressRetireWarningMessage);

BethanyZhou (changes requested):
LGTM, left one message.

src/Common/AzurePSCmdlet.cs
erich-wang: Good suggestion. However Environment.GetEnvironmentVariable("SuppressAzureRmModulesRetiringWarning") may throw exception per https://docs.microsoft.com/en-us/dotnet/api/system.environment.getenvironmentvariable?view=net-5.0, we need to use try catch anyway...

BethanyZhou: Make sense. Thank your for clarification.

PR #262, opened 2021-05-14
add survey helper for cmdlet base

VeryEarly:


src/Common/AzurePSCmdlet.cs
```diff
@@ -434,6 +438,37 @@ protected bool IsVerbose()
             return verbose;
         }
 
+        protected void WriteSurvey()
+        {
+            HostInformationMessage msg = new HostInformationMessage()
+            {
+                Message = string.Format("Survey: How was your experience using {0}.{1}?`\nRun ", ModuleName, ModuleVersion),
```
dingmeng-xue, line 445: Since we don't raise survey per module. Let's use Azure PowerShell for the time being.

src/Common/AzurePSCmdlet.cs
```diff
@@ -434,6 +438,37 @@ protected bool IsVerbose()
...
+                NoNewLine = true
+            };
+            HostInformationMessage msg2;
+            try
+            {
+                msg2 = new HostInformationMessage()
+                {
+                    Message = "Open-AzSurveyLink",
+                    NoNewLine = true,
+                    ForegroundColor = (ConsoleColor)Host.PrivateData.Properties.Match("ProgressBackgroundColor").SingleOrDefault().Value
```
dingmeng-xue, line 455: Get color once in one process. If there is exception, fallback to color of warning message?
After it, you don't need to try-catch in each cmdlet.

src/Common/Survey/ModuleInfo.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        internal ModuleInfo() { }
+
+        internal bool IsDeprecatedVersion(int version)
+        {
+            string str = Convert.ToString(DeprecatedVersions, 2);
+            return version == 0 ? false : str[str.Length - version].Equals('1');
+        }
+
+        internal void Deprecate(string date)
```
dingmeng-xue, line 50: What's the meaning of deprecation? I need you to describe the structure of json file.

src/Common/Survey/ModuleInfo.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            DeprecateVersion();
+            FirstActiveDate = date;
+            LastActiveDate = date;
+            Count = 0;
+        }
+
+        private void DeprecateVersion()
```
dingmeng-xue, line 58: Dotnet has object to handle version. You don't need to calculate it by your code. You just need to exclude suffix (version label) before converting it to version object.

src/Common/Survey/ScheduleInfo.cs
```diff
@@ -0,0 +1,21 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.WindowsAzure.Commands.Common.Survey
+{
+    public class ScheduleInfo
+    {
+        [JsonProperty(PropertyName = "lock")]
+        internal string Lock { get; set; }
```
dingmeng-xue, line 11: this class should map to AzureRmSurvey. its name should reflect it.

src/Common/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,318 @@
...
+            FlushCount = 0;
+        }
+
+        private static SurveyHelper _instance
+        {
+            get
+            {
+                if (_instance == null)
+                {
+                    return new SurveyHelper();
```
dingmeng-xue, line 36: _instance=new SurveyHelper();

src/Common/Survey/ModuleInfo.cs
```diff
@@ -0,0 +1,85 @@
+﻿using Newtonsoft.Json;
```
erich-wang, line 1: Please add license header

src/Common/AzurePSCmdlet.cs
```diff
@@ -434,6 +438,37 @@ protected bool IsVerbose()
             return verbose;
         }
 
+        protected void WriteSurvey()
+        {
+            HostInformationMessage msg = new HostInformationMessage()
+            {
+                Message = string.Format("Survey: How was your experience using Azure PowerShell?`\nRun ", ModuleName, ModuleVersion),
```
isra-fel, line 445:
```suggestion
                Message = "Survey: How was your experience using Azure PowerShell?`\nRun "
```
and use resource string please

src/Common/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,335 @@
...
+        private const int _lockExpiredDays = 30;
+        private const int _surveyTriggerCount = 3;
+        private const int _flushFrequecy = 5;
+        private const int _delayForSecondPrompt = 2;
+        private const int _delayForThirdPrompt = 5;
+
+        private SurveyHelper()
+        {
+            CurrentDate = DateTime.UtcNow.ToString("yyyy'-'MM'-'dd");
+            IgnoreSchedule = "Disabled".Equals(Environment.GetEnvironmentVariable(AzurePowerShell.AzurePSInterceptSurvey));
```
erich-wang, line 26: case sensitive?

src/Common/AzurePSCmdlet.cs
```diff
@@ -434,6 +438,37 @@ protected bool IsVerbose()
             return verbose;
         }
 
+        protected void WriteSurvey()
+        {
+            HostInformationMessage msg = new HostInformationMessage()
```
isra-fel, line 443: We can use more meaningful names
```suggestion
            HostInformationMessage surveyIntroMsg = new HostInformationMessage()
```

src/Common/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,335 @@
...
+
+        private DateTime LastPromptDate { get; set; }
+
+        private IDictionary<string, ModuleInfo> InternalMap { get; set; }
+
+        private bool IgnoreSchedule;
+
+        private readonly string CurrentDate;
+
+        private string ActiveModuleName;
```
erich-wang, line 56: Suggest to move all fields at the beginning of class

src/Common/AzurePowerShell.cs
```diff
@@ -51,5 +51,9 @@ public class AzurePowerShell
 #endif
             Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData),
             "Windows Azure PowerShell");
+
+        public static string SurveyScheduleInfoDirectory = Path.Combine(ProfileDirectory, "AzureRmSurvey.json");
```
isra-fel, line 55: nit: this is a file
```suggestion
        public static string SurveyScheduleInfoFile = Path.Combine(ProfileDirectory, "AzureRmSurvey.json");
```

src/Common/Survey/ModuleInfo.cs
```diff
@@ -0,0 +1,44 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.WindowsAzure.Commands.Common.Survey
+{
+    public class ModuleInfo
```
isra-fel, line 8: I suppose the class can be internal?

src/Common/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,335 @@
...
+            ActiveModuleName = null;
+        }
+
+        private static SurveyHelper _instance
+        {
+            get
+            {
+                return _instance ?? new SurveyHelper();
+            }
+        }
```
erich-wang, line 40: Endless loop?

src/Common/Survey/ModuleInfo.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        [JsonProperty(PropertyName = "name")]
+        internal string Name { get; set; }
+
+        [JsonProperty(PropertyName = "version")]
+        internal int Version { get; set; }
+
+        [JsonProperty(PropertyName = "count")]
+        internal int Count { get; set; }
```
isra-fel, line 17: Could use some explaination
```suggestion
        /// <summary>Count of ???</summary>
        [JsonProperty(PropertyName = "count")]
        internal int Count { get; set; }
```

src/Common/Survey/ScheduleInfo.cs
```diff
@@ -0,0 +1,21 @@
...
+        [JsonProperty(PropertyName = "lastPromptDate")]
+        internal string LastPromptDate { get; set; }
+
+        [JsonProperty(PropertyName = "modules")]
+        internal IList<ModuleInfo> Modules { get; set; }
+
+        [JsonProperty(PropertyName = "interceptTriggered")]
+        internal int InterceptTriggered { get; set; }
+
+        public void MergeScheduleInfo(ScheduleInfo externalScheduleInfo) { }
```
isra-fel, line 19: Can this be removed?

erich-wang (changes requested):

src/Common/Survey/SurveyHelper.cs
isra-fel, line 40: `_instance` is never set, so every time a new instance will be created.
```suggestion
        private static SurveyHelper _instance { get; set; }
        public static SurveyHelper GetInstance()
        {
            if (_instance == null)
            {
                _instance = new SurveyHelper();
            }
            return _instance;
        }
```

src/Common/Survey/SurveyHelper.cs
isra-fel, line 26: We might want to be more tolerant to casing?
```suggestion
            IgnoreSchedule = "Disabled".Equals(Environment.GetEnvironmentVariable(AzurePowerShell.AzurePSInterceptSurvey, StringComparsion.xxxIgnoreCase));
```

src/Common/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,335 @@
...
+
+        private int FlushCount;
+
+        private int InterceptTriggered;
+
+        private static string SurveySchedulePath = AzurePowerShell.SurveyScheduleInfoDirectory;
+
+        private DateTime LastPromptDate { get; set; }
+
+        private IDictionary<string, ModuleInfo> InternalMap { get; set; }
```
isra-fel, line 50:
```suggestion
        private IDictionary<string, ModuleInfo> _modules { get; set; }
```

src/Common/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,335 @@
...
+
+        private static SurveyHelper _instance
+        {
+            get
+            {
+                return _instance ?? new SurveyHelper();
+            }
+        }
+
+        private int FlushCount;
```
isra-fel, line 42: private fields are named `_camelCase`

src/Common/Survey/SurveyHelper.cs
erich-wang, line 50: Suggest to use its real type ConcurrentDictionary

src/Common/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,335 @@
...
+        {
+            if (IgnoreSchedule)
+            {
+                return false;
+            }
+
+            int majorVersion = Version.Parse(moduleVersion).Major;
+
+            DateTime today = Convert.ToDateTime(CurrentDate);
+            int delay = InterceptTriggered == 1 ? today.CompareTo(LastPromptDate.AddDays(_delayForSecondPrompt)) : (InterceptTriggered == 2 ? today.CompareTo(LastPromptDate.AddDays(_delayForThirdPrompt)) : -1);
```
erich-wang, line 73: When will InterceptTriggered be assigned as 1?

isra-fel (changes requested):

src/Common/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,336 @@
...
+        private string ActiveModuleName;
+
+        public static SurveyHelper GetInstance()
+        {
+            return _instance;
+        }
+
+        public bool ShouldPropmtSurvey(string moduleName, string moduleVersion)
+        {
+            if (IgnoreSchedule)
```
VeryEarly, line 65: get environment variable

src/Common/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,336 @@
...
+                    LastActiveDate = CurrentDate,
+                    DeprecatedVersions = 0,
+                    Enabled = true
+                };
+            }
+
+            ModuleInfo cur = InternalMap[moduleName];
+
+            //LastPromptDate.CompareTo(DateTime.MinValue) > 0 means survey is locked, otherwise lock free
+            if (LastPromptDate.CompareTo(DateTime.MinValue) > 0 && Convert.ToDateTime(CurrentDate).CompareTo(LastPromptDate.AddDays(_lockExpiredDays)) > 0)
```
VeryEarly, line 104: need to be fixed

src/Common/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,336 @@
...
+        private bool ReadFromStream(out ScheduleInfo info)
+        {
+            string name = null;
+            bool prompt = false;
+            return ReadFromStream(name, out prompt, out info);
+        }
+
+        private bool ReadFromStream(string moduleName, out bool prompt, out ScheduleInfo info)
+        {
+            FileStream fs = new FileStream(SurveySchedulePath, FileMode.Open, FileAccess.Read, FileShare.None);
```
VeryEarly, line 216: move this line into try block

src/Common/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,336 @@
...
+            {
+                Interlocked.Exchange(ref FlushCount, 0);
+            }
+            else
+            {
+                Interlocked.Add(ref FlushCount, 1);
+            }
+        }
+
+        private async void TryFlush()
```
VeryEarly, line 318: rename TryFlushAsync

src/Common/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,336 @@
...
+            CurrentDate = DateTime.UtcNow.ToString("yyyy'-'MM'-'dd");
+            IgnoreSchedule = "Disabled".Equals(Environment.GetEnvironmentVariable(AzurePowerShell.AzurePSInterceptSurvey));
+            LastPromptDate = DateTime.MinValue;
+            InternalMap = new ConcurrentDictionary<string, ModuleInfo>();
+            FlushCount = 0;
+            InterceptTriggered = 0;
+            ActiveModuleName = null;
+        }
+
+        private static SurveyHelper _instance
```
VeryEarly, line 34: uppercase

src/Share/Constants.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public class Constants
+    {
+        public static string ProfileDirectory = Path.Combine(
+            Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),
+            ".Azure");
+
+        public static string SurveyScheduleInfoFile = Path.Combine(ProfileDirectory, "AzureRmSurvey.json");
+
+        public const string AzurePSInterceptSurvey = "Azure_PS_Intercept_Survey";
```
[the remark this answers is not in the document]
dingmeng-xue: Agree that we should not expose those variables if it is not used by others.

src/Common/AzurePowerShell.cs
```diff
@@ -51,5 +51,9 @@ public class AzurePowerShell
 #endif
             Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData),
             "Windows Azure PowerShell");
+
+        public static string SurveyScheduleInfoFile = Path.Combine(ProfileDirectory, "AzureRmSurvey.json");
```
dingmeng-xue, line 55: It should be removed if this logic is controlled by share lib.

src/Common/MetricHelper.cs
```diff
@@ -530,6 +531,7 @@ public class AzurePSQoSEvent
     public string SessionId { get; set; }
     public string SubscriptionId { get; set; }
     public string TenantId { get; set; }
+    public bool SurveyPrompted { get; set; }
```
dingmeng-xue, line 534: it should be nullable bool?. we need to add it into telemetry only when its value is true.

src/Share/Survey/ScheduleInfo.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using Newtonsoft.Json;
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.PowerShell.Share.Survey
+{
+    internal class ScheduleInfo
+    {
+        [JsonProperty(PropertyName = "lastPromptDate")]
+        internal string LastPromptDate { get; set; }
```
dingmeng-xue, line 23: Predictor may have higher prompt frequency. Can we allow it set its value? If yes, LastPromptDate should be in moduleinfo of predictor.

src/Share/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,353 @@
...
+        private static string SurveySchedulePath = Constants.SurveyScheduleInfoFile;
+        private static string Predictor = Constants.Predictor;
+
+        private DateTime LastPromptDate { get; set; }
+
+        private ConcurrentDictionary<string, ModuleInfo> Modules { get; set; }
+
+        private bool IgnoreSchedule;
+
+        private bool IsDisabledFromEnv => "Disabled".Equals(Environment.GetEnvironmentVariable(Constants.AzurePSInterceptSurvey), StringComparison.OrdinalIgnoreCase);
```
dingmeng-xue, line 51: Can we support "false" too?

src/Share/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,353 @@
...
+        }
+
+        private bool ReadFromStream()
+        {
+            StreamReader sr = null;
+            try
+            {
+                if (File.Exists(SurveySchedulePath))
+                {
+                    sr = new StreamReader(new FileStream(SurveySchedulePath, FileMode.Open, FileAccess.Read, FileShare.None));
```
dingmeng-xue, line 259: Please use this API https://docs.microsoft.com/en-us/dotnet/api/system.io.file.readalltext?view=net-5.0
You don't need to read it via Stream.

src/Share/Survey/ScheduleInfo.cs
VeryEarly: AzPredictor has its own logic at line193

src/Common/MetricHelper.cs
VeryEarly: no need for three states, add logic to add it only when value is true

src/Share/SharedVariable.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    /// <summary>
+    /// Static class containing variables shared across assemblies
+    /// </summary>
+    public static class SharedVariable
+    {
+        /// <summary>
+        /// Id should be populated by Az.Tools.Predictor and recorded by AzPS cmdlet telemetry. Once it is recorded, value of id will be cleaned.
+        /// </summary>
+        public static string PredictorCorrelationId { get; set; }
```
kceiw, line 25: This can be read/written in multiple threads. As least, the underlying field should be [volatile](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/volatile)

src/Share/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,360 @@
...
+                                        || "False".Equals(Environment.GetEnvironmentVariable(AzurePSInterceptSurvey), StringComparison.OrdinalIgnoreCase);
+
+        private readonly string CurrentDate;
+
+        private readonly DateTime Today;
+
+        private SurveyHelper()
+        {
+            CurrentDate = DateTime.UtcNow.ToString("yyyy-MM-dd");
+            Today = Convert.ToDateTime(CurrentDate);
```
kceiw, line 66: `DateTime.Today` returns today.
`CurrentDate` and `Today` are the same data but in different format. Do we need to keep the two?

src/Share/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,360 @@
...
+        }
+
+        private bool ReadFromStream()
+        {
+            StreamReader sr = null;
+            try
+            {
+                if (File.Exists(SurveyScheduleInfoFile))
+                {
+                    sr = new StreamReader(new FileStream(SurveyScheduleInfoFile, FileMode.Open, FileAccess.Read, FileShare.None));
```
kceiw, line 259: Should we allow multiple read by using `FileShare.Read`? Otherwise once one module is reading the file, the other modules won't be able to read it.

src/Share/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,360 @@
...
+            catch (Exception e)
+            {
+                if (e is UnauthorizedAccessException)
+                {
+                    this.IgnoreSchedule = true;
+                }
+                //deserialize failed, means content of file is incorrect, make file empty
+                if (e is JsonException)
+                {
+                    if (sr != null)
```
kceiw, line 272: `sr.Dispose` is done in `finally` already. This is not necessary.

src/Share/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,360 @@
...
+                FlushWithWait();
+            });
+        }
+
+        private void FlushWithWait()
+        {
+            int beforeExchange = Interlocked.CompareExchange(ref FlushCount, 0, FlushFrequecy);
+            if (beforeExchange < FlushFrequecy)
+            {
+                Interlocked.Add(ref FlushCount, 1);
```
kceiw, line 345: `Interlocked.Increment` is doing "+1". It won't be worse than `Add` performance wise.

src/Share/Survey/SurveyHelper.cs
```diff
@@ -0,0 +1,360 @@
...
+            }
+            return true;
+        }
+
+        private bool WriteToStream(string info)
+        {
+            StreamWriter sw = null;
+            try
+            {
+                sw = new StreamWriter(new FileStream(SurveyScheduleInfoFile, FileMode.Create, FileAccess.Write, FileShare.None));
```
kceiw, line 295: The read and write can be on different threads. With this, the following write will fail once a thread is reading/writing. This may not be a big deal. Just want to bring it to your attentions.

src/Share/Survey/SurveyHelper.cs
VeryEarly: these two variables will be used quite often at runtime

src/Share/Survey/SurveyHelper.cs
VeryEarly: You right, read should not block each other.

src/Share/Survey/SurveyHelper.cs
VeryEarly: Our strategy is to fail other read/write while one thread is writing.
Failures of read/write were handled and acceptable.

src/Share/Survey/SurveyHelper.cs
VeryEarly: this is block is a fallback when serialization failed, it will invoke a new write stream, thus the current read stream has to be disposed.

dingmeng-xue (approved):

dingmeng-xue (dismissed):

src/Share/SharedVariable.cs
VeryEarly: added

src/Share/Survey/SurveyHelper.cs
VeryEarly: done

PR #269, opened 2021-06-25
Generate subscripition client of version 2021-01-01

msJinLei:
This PR generates subscripition client of version 2021-01-01
It has been tested with azure-powershell


src/ResourceManager/Version2021_01_01/Customized/SubscriptionExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// ----------------------------------------------------------------------------------
+
+using System.Linq;
+using Microsoft.Azure.Commands.Common.Authentication.Abstractions;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Management.ResourceManager.Version2021_01_01.Models.Utilities
```
isra-fel, line 19: I think the namespace should be `Microsoft.Azure.Commands.ResourceManager.Version2021_01_01.Customized`, just like `SubscriptionClientExtension.cs`

src/ResourceManager/Version2021_01_01/Customized/TenantIdDescriptionExtensions.cs
```diff
@@ -0,0 +1,66 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// ----------------------------------------------------------------------------------
+
+using System.Linq;
+using Microsoft.Azure.Commands.Common.Authentication;
+using Microsoft.Azure.Commands.Common.Authentication.Abstractions;
+using Microsoft.Azure.Commands.ResourceManager.Common;
+
+namespace Microsoft.Azure.Management.ResourceManager.Version2021_01_01.Models.Utilities
```
isra-fel, line 20: Here too

src/ResourceManager/Version2021_01_01/readme.md
```diff
@@ -0,0 +1,41 @@
...
+namespace: Microsoft.Azure.Management.ResourceManager.Version2021_01_01
+
+#directive:
+# - remove-operation:
+#   - Operations_List
+#   - Subscriptions_ListLocations
+#   - Subscriptions_Get
+#   - Subscriptions_List
+#   - Tenants_List
+#   - checkResourceName
```
isra-fel, line 40: Does this mean we want to keep all the operations?

isra-fel (commented):
Have you tested this new version?

src/ResourceManager/Version2021_01_01/readme.md
msJinLei: yes, the operations are not so many. I would like to keep them all.

src/ResourceManager/Version2021_01_01/Customized/SubscriptionExtensions.cs
msJinLei: To keep the consistent with the existing codes, we prefer to use Microsoft.Azure.Management.ResourceManager.Version2021_01_01.Models.Utilities.

isra-fel (approved):
LGTM

src/ResourceManager/Version2021_01_01/readme.md
msJinLei: check subscription client with previous version and find all the operations are kept.

PR #276, opened 2021-07-05
Upgrade Microsoft.ApplicationInsights from 2.4.0 to 2.12.0

msJinLei:
- Upgrade Microsoft.ApplicationInsights from 2.4.0 to 2.12.0

### DIFF
--- a/src/Common/MetricHelper.cs
+++ b/src/Common/MetricHelper.cs
@@ -127,7 +127,7 @@ namespace Microsoft.WindowsAzure.Commands.Common
 #if DEBUG
             if (TestMockSupport.RunningMocked)
             {
-                TelemetryConfiguration.Active.DisableTelemetry = true;
+                telemetryConfiguration.DisableTelemetry = true;
             }
 #endif
         }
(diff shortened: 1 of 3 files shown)


src/Common/MetricHelper.cs
```diff
@@ -425,7 +425,7 @@ private void PopulatePropertiesFromQos(AzurePSQoSEvent qos, IDictionary<string,
         private static string[] exceptionTrackAcceptModuleList = { "Az.Accounts", "Az.Compute", "Az.AKS", "Az.ContainerRegistry" };
         private static string[] exceptionTrackAcceptCmdletList = { "Get-AzKeyVaultSecret", "Get-AzKeyVaultCert" };
 
-        private static string ConvertFrameToString(StackFrame frame)
+        private static string ConvertFrameToString(System.Diagnostics.StackFrame frame)
```
isra-fel, line 428: Change on this line doesn't seem necessary

src/Common/MetricHelper.cs
```diff
@@ -127,7 +127,7 @@ public MetricHelper(INetworkHelper network)
 #if DEBUG
             if (TestMockSupport.RunningMocked)
             {
-                TelemetryConfiguration.Active.DisableTelemetry = true;
+                telemetryConfiguration.DisableTelemetry = true;
```
isra-fel, line 130: The active telemetry configuration being used may not be `telemetryConfiguration`, so I suppose `TelemetryConfiguration.Active` makes more sense here

isra-fel (changes requested):
~~Have you tested it out with powershell modules?~~
Jin Lei has tested locally and judged from the traces the telemetry data was sent to Azure

src/Common/MetricHelper.cs
msJinLei: StackFrame also defined in new version of Application Insights namespace and so I add prefix here.

src/Common/MetricHelper.cs
msJinLei: Using TelemetryConfiguration.Active is not recommended on .NET Core. See https://github.com/microsoft/ApplicationInsights-dotnet/issues/1152 for more details

isra-fel (approved):

PR #277, opened 2021-07-12
Fix access error when subscripiton has no Tags property

msJinLei:
Related issue: https://github.com/Azure/azure-powershell/issues/15425


isra-fel (approved):
LGTM

dingmeng-xue (approved):

PR #278, opened 2021-07-26
Get correct PSVersion from PowerShell runtime

dingmeng-xue:
PSVersion previously is fetch via $Host.Version. After discussion with PowerShell team, the correct way is $Host.Runspace.Version. So, it changes below codes:
* Get correct PowerShell version via `$Host.Runspace.Version` in PowerShell scripts
* Keep host name and version into new telemetry properties
* Add additional information into telemetry debug message
* Clean deprecated code.


isra-fel (approved):
LGTM

VeryEarly (approved):

PR #282, opened 2021-08-06
Enable custom subscription ID by attribute

isra-fel:
### Problem

autorest-based modules support `-Subscription` parameter to operate on non-default subscription, but there's no universal way to enable it for sdk-based ones.

### Solution

A `SupportsSubscriptionId` attribute is introduced. When applied to a cmdlet (or a cmdlet base class), the cmdlet(s) will have a `-SubscriptionId` parameter, implemented by `IDynamicParameter`. And during its lifecycle, the default profile and context will be cloned, the subscription will be replaced by what user specifies. Therefore when creeating SDK client or sending request, the input subscription ID will be used.

### DIFF
--- a/src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
+++ b/src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
@@ -23,9 +23,11 @@ using Microsoft.Azure.Management.Internal.Resources;
 using Microsoft.Azure.Management.Internal.Resources.Utilities.Models;
 using Microsoft.Rest;
 using Microsoft.WindowsAzure.Commands.Common;
+using Microsoft.WindowsAzure.Commands.Common.Attributes;
 using Microsoft.WindowsAzure.Commands.Utilities.Common;
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Linq;
 using System.Management.Automation;
 using System.Net.Http.Headers;
@@ -37,7 +39,7 @@ namespace Microsoft.Azure.Commands.ResourceManager.Common
     /// <summary>
     /// Represents base class for Resource Manager cmdlets
     /// </summary>
-    public abstract class AzureRMCmdlet : AzurePSCmdlet
+    public abstract class AzureRMCmdlet : AzurePSCmdlet, IDynamicParameters
     {
         protected ServiceClientTracingInterceptor _serviceClientTracingInterceptor;
         IAzureContextContainer _profile;
@@ -50,6 +52,7 @@ namespace Microsoft.Azure.Commands.ResourceManager.Common
         public const string WriteVerboseKey = "WriteVerbose";
         public const string WriteWarningKey = "WriteWarning";
         public const string EnqueueDebugKey = "EnqueueDebug";
+        private const string SubscriptionIdParameter = "SubscriptionId";
 
         /// <summary>
         /// Creates new instance from AzureRMCmdlet and add the RPRegistration handler.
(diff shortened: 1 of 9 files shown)


src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
```diff
@@ -546,5 +592,27 @@ private void EnqueueDebugSender(object sender, StreamEventArgs args)
...
+        public object GetDynamicParameters()
+        {
+            var parameters = new RuntimeDefinedParameterDictionary();
+
+            // add `-SubscriptionId` if the cmdlet has [SupportsSubscriptionId] attribute
+            if (GetType().IsDefined(typeof(SupportsSubscriptionIdAttribute), true))
+            {
+                const string helpMessage = @"The ID of the subscription.
+By default, cmdlets are executed in the subscription that is set in the current context. If the user specifies another subscription, the current cmdlet is executed in the subscription specified by the user.
+Overriding subscriptions only take effect during the lifecycle of the current cmdlet. It does not change the subscription in the context, and does not affect subsequent cmdlets.";
```
isra-fel, line 605: Cannot use resource string because when VS regenerates `Resources.Designer.cs` all public becomes internal, but actually the resource strings are used by az.accounts, and nobody want to manually edit that file, so

isra-fel (commented):
Need to add some test cases

src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
```diff
@@ -67,23 +70,66 @@ public IAzureContextContainer DefaultProfile
...
+
+        private IAzureContextContainer CloneProfileAndModifyContext()
+        {
+            // going to modify default context, so only shallow copying other stuff
+            var clonedProfile = GetDefaultProfile().ShallowCopy();
+            clonedProfile.DefaultContext = clonedProfile.DefaultContext.DeepCopy();
+
+            if (MyInvocation.BoundParameters.TryGetValue(SubscriptionIdParameter, out var overriddenSub))
+            {
+                var matchingSub = clonedProfile.Subscriptions.FirstOrDefault(sub => sub.GetId().Equals(new Guid(overriddenSub as string)));
```
erich-wang, line 111: How about put `new Guid(overriddenSub as string)` ahead for better performance?

erich-wang: Should we raise meaningful error message if provided sub id is illegal, e.g. "123"?

src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
```diff
@@ -67,23 +70,66 @@ public IAzureContextContainer DefaultProfile
...
+                if (matchingSub != null)
```
erich-wang, line 112: Should we tell customer the provided sub id is not found if matchingSub is null?

src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
erich-wang: Actually Resources.resx allows us to specify default access modifier, not sure if this would help.

src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
```diff
@@ -37,7 +39,7 @@ namespace Microsoft.Azure.Commands.ResourceManager.Common
     /// <summary>
     /// Represents base class for Resource Manager cmdlets
     /// </summary>
-    public abstract class AzureRMCmdlet : AzurePSCmdlet
+    public abstract class AzureRMCmdlet : AzurePSCmdlet, IDynamicParameters
```
erich-wang, line 42: Does it affect the cmdlets that have already implemented `IDynamicParameters`?

erich-wang (changes requested):

src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
isra-fel: I tested with `Add-AzIotHubModule -AuthMethod x509_thumbprint -PrimaryThumbprint` and it worked well on Windows PowerShell and powershell 7.
Supposedly the reason there's no error about conflicting methods is because [CS0109](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0109) is a **compiler warning**.

msJinLei (approved):

erich-wang (approved):

PR #288, opened 2021-09-18
Add endpoint for MSGraph

wyunchi-ms:


dingmeng-xue (dismissed):

src/Authentication.Abstractions/Interfaces/IAzureEnvironment.cs
```diff
@@ -136,5 +136,15 @@ public interface IAzureEnvironment : IExtensibleModel
...
+
+        /// <summary>
+        /// The MSGraph endpoint
+        /// </summary>
+        string MicrosoftGraphUrl { get; set; }
+
+        /// <summary>
+        /// The MSGraph Resource Url endpoint
+        /// </summary>
+        string MicrosoftGraphEndpointResourceId { get; set; }
```
isra-fel, line 148: When adding new endpoints (resource IDs), we typically add them as extended properties of IAzureEnvironment, so that the interface doesn't need to change.
You can refer to this example: https://github.com/Azure/azure-powershell-common/pull/227/files

isra-fel (changes requested):

src/Authentication.Abstractions/Interfaces/IAzureEnvironment.cs
wyunchi-ms: Resolved

src/Authentication.Abstractions/AzureEnvironment.cs
```diff
@@ -475,6 +483,16 @@ public AzureEnvironment(IAzureEnvironment other)
...
+        /// <summary>
+        /// The MSGraph endpoint
+        /// </summary>
+        public string MicrosoftGraphUrl { get; set; }
+
+        /// <summary>
+        /// The token audience need for tokens that target MSGraph endpoints
+        /// </summary>
+        public string MicrosoftGraphEndpointResourceId { get; set; }
+
```
isra-fel, line 495: I suppose this is no longer used. We should consider removing it.

isra-fel (changes requested):
One comment. Others LGTM

src/Authentication.Abstractions/AzureEnvironment.cs
wyunchi-ms: Resolved

isra-fel (approved):

msJinLei (approved):

PR #299, opened 2021-11-10
Add module version to useragent

dingmeng-xue:
For https://github.com/Azure/azure-powershell/issues/16291

Fetch module name and module version in  `BeginProcessing()` instead of `InitializeQosEvent()` because those 2 information are required by telemetry and useragent both.

Currently, we append 3 product information into useragent, AzurePowerShell, PowerShell, and module now. Azure PowerShell and PowerShell are constant in one PowerShell process but module information needs to be removed after cmdlet execution. But remove method `AzureSession.Instance.ClientFactory.RemoveUserAgent(ModuleName);` is not safe when cmdlet execution is cancelled. It will be addressed in `azure-powershell\src\Accounts\Authentication\Factories\ClientFactory.cs`

### DIFF
--- a/src/Common/AzurePSCmdlet.cs
+++ b/src/Common/AzurePSCmdlet.cs
@@ -110,16 +110,18 @@ namespace Microsoft.WindowsAzure.Commands.Utilities.Common
             }
         }
 
+        private string _moduleName;
+        private string _moduleVersion;
         /// <summary>
         /// Gets the PowerShell module name used for user agent header.
         /// By default uses "Azure PowerShell"
         /// </summary>
-        protected virtual string ModuleName { get { return "AzurePowershell"; } }
+        protected virtual string ModuleName { get { return _moduleName; } set { this._moduleName = value; } }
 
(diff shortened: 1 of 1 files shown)


src/Common/AzurePSCmdlet.cs
```diff
@@ -110,16 +110,18 @@ protected string PSVersion
...
         }
 
+        private string _moduleName;
+        private string _moduleVersion;
         /// <summary>
         /// Gets the PowerShell module name used for user agent header.
         /// By default uses "Azure PowerShell"
         /// </summary>
-        protected virtual string ModuleName { get { return "AzurePowershell"; } }
+        protected virtual string ModuleName { get { return _moduleName; } set { this._moduleName = value; } }
```
isra-fel, line 119: If there's no special logic in getter and setter, this is equivalant to
```suggestion
        protected virtual string ModuleName { get; set; }
```

BethanyZhou (approved):

PR #305, opened 2021-12-28
Fixed the bug of cmdlet fails when -DefaultProfile is set to service …

msJinLei:
…principal login context

To fix https://github.com/Azure/azure-powershell/issues/16617
copy ServicePrincipalSecret and CertificatePassword from AzureRmProfileProvider.Instance.Profile.Accounts
if current account doesn't have such information

### DIFF
--- a/src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
+++ b/src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
@@ -86,7 +86,56 @@ namespace Microsoft.Azure.Commands.ResourceManager.Common
             set
             {
                 _profile = value;
+                var accountMap = CreateAccountMap(AzureRmProfileProvider.Instance.Profile.Accounts);
+                foreach (var currentAccount in _profile.Accounts)
+                {
+                    if (accountMap.ContainsKey(currentAccount.Id))
+                    {
+                        var accList = accountMap[currentAccount.Id];
+                        foreach (var acc in accList)
+                        {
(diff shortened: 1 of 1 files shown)


src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
```diff
@@ -86,7 +86,56 @@ public IAzureContextContainer DefaultProfile
             set
             {
                 _profile = value;
+                var accountMap = CreateAccountMap(AzureRmProfileProvider.Instance.Profile.Accounts);
+                foreach (var currentAccount in _profile.Accounts)
+                {
+                    if (accountMap.ContainsKey(currentAccount.Id))
```
dingmeng-xue, line 92:
```suggestion
                foreach (var currentAccount in _profile.Accounts)
                {
                    if (accountMap.ContainsKey(currentAccount.Id))
```
You can use `public bool TryGetValue (TKey key, out TValue value);`

src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
```diff
@@ -86,7 +86,56 @@ public IAzureContextContainer DefaultProfile
             set
             {
                 _profile = value;
+                var accountMap = CreateAccountMap(AzureRmProfileProvider.Instance.Profile.Accounts);
```
dingmeng-xue, line 89: Please add comment to here. We need to explain why we need to load secret from context.

wyunchi-ms (approved):

dingmeng-xue (approved):

PR #314, opened 2022-03-23
Use config

isra-fel:


src/ResourceManager/Properties/Resources.resx
```diff
@@ -230,4 +230,7 @@ Overriding subscriptions only take effect during the lifecycle of the current cm
   <data name="CustomSubscriptionNotFound" xml:space="preserve">
     <value>Could not found subscription '{0}' in context. Please run 'Connect-AzAccount' with correct user.</value>
   </data>
+  <data name="SubscriptionNotFouldPleaseCheckConfig" xml:space="preserve">
```
isra-fel, line 233: Move this message to Az.Accounts

PR #316, opened 2022-04-13
Upgrade the version of Microsoft.Rest.ClientRuntime.Azure

dingmeng-xue:
Just upgrade the version of Microsoft.Rest.ClientRuntime.Azure. We don't need to release a new version only for this change.


isra-fel (approved):
LGTM

BethanyZhou (approved):
LGTM

PR #318, opened 2022-05-15
EnableDataCollection and DisplayBreakingChangeWarnings configs

isra-fel:
Includes the definition of the following configs:
- DisplayBreakingChangeWarning
- EnableDataCollection
- EnableSurvey (delayed)

And modification to config definition model:
- Made `EnvironmentVariable` protected because it does not need to be public
- Exposed the ability to customize how environment variables are parsed to the config value by `ConfigDefinition.ParseFromEnvironmentVariables()`


BethanyZhou (approved):
LGTM

wyunchi-ms (approved):

PR #319, opened 2022-05-31
Fixed a telemtry related issue in multi-thread

dolauli:
The root cause of the issue is as listed in [here](https://docs.microsoft.com/en-us/dotnet/api/system.threadstaticattribute?redirectedfrom=MSDN&view=net-6.0). 
`Initialization occurs only once, when the class constructor executes, and therefore affects only one thread.`

### DIFF
--- a/src/Common/MetricHelper.cs
+++ b/src/Common/MetricHelper.cs
@@ -46,9 +46,9 @@ namespace Microsoft.WindowsAzure.Commands.Common
         /// Set _telemetryId and _internalCalledCmdlets as thread local since we need an instance of them per thread.
         /// </summary>
         [ThreadStatic]
-        private static string _telemetryId = "";
+        private static string _telemetryId;
         [ThreadStatic]
-        private static string _internalCalledCmdlets = "";
+        private static string _internalCalledCmdlets;
 
         public static string TelemetryId { get => _telemetryId ?? ""; set { _telemetryId = value; } }
 
--- a/src/Common/MetricHelper.cs
+++ b/src/Common/MetricHelper.cs
@@ -54,7 +54,7 @@ namespace Microsoft.WindowsAzure.Commands.Common
 
         public static string InternalCalledCmdlets { get => _internalCalledCmdlets ?? ""; set { _internalCalledCmdlets = value; } }
 
-        public static bool IsCalledByUser() { return string.IsNullOrEmpty(_telemetryId) ? true : false; }
+        public static bool IsCalledByUser() { return string.IsNullOrEmpty(_telemetryId); }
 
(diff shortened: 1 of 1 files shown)


src/Common/MetricHelper.cs
```diff
@@ -50,13 +50,13 @@ public class MetricHelper
...
         private static string _internalCalledCmdlets = "";
 
-        public static string TelemetryId { get=> _telemetryId; set { _telemetryId = value; } }
+        public static string TelemetryId { get => _telemetryId ?? ""; set { _telemetryId = value; } }
 
-        public static string InternalCalledCmdlets { get => _internalCalledCmdlets; set { _internalCalledCmdlets = value; } }
+        public static string InternalCalledCmdlets { get => _internalCalledCmdlets ?? ""; set { _internalCalledCmdlets = value; } }
 
-        public static bool IsCalledByUser() { return _telemetryId == "" ? true : false; }
+        public static bool IsCalledByUser() { return string.IsNullOrEmpty(_telemetryId) ? true : false; }
```
isra-fel, line 57:
```suggestion
        public static bool IsCalledByUser() { return string.IsNullOrEmpty(_telemetryId); }
```

src/Common/MetricHelper.cs
```diff
@@ -50,13 +50,13 @@ public class MetricHelper
         [ThreadStatic]
         private static string _internalCalledCmdlets = "";
 
-        public static string TelemetryId { get=> _telemetryId; set { _telemetryId = value; } }
+        public static string TelemetryId { get => _telemetryId ?? ""; set { _telemetryId = value; } }
```
isra-fel, line 53: To my understanding, if _internalCalledCmdlets is not initialized, its default value would be null. But we prefer empty string than null, that's the reason you made this change, right?

dolauli: > To my understanding, if _internalCalledCmdlets is not initialized, its default value would be null. But we prefer empty string than null, that's the reason you made this change, right?

You are right. The issue here is I used to think assignment in line#51 will happen in all the threads, but the assignment only happen in the first thread.

src/Common/MetricHelper.cs
```diff
@@ -50,13 +50,13 @@ public class MetricHelper
         [ThreadStatic]
         private static string _internalCalledCmdlets = "";
```
vidai-msft, line 51: Please remove the field initialization to string type as by default it should be null in order to avoid any confusion when some one unfamiliar with this attribute reads the code in the future.

vidai-msft (changes requested):

vidai-msft (approved):

isra-fel (approved):

PR #329, opened 2022-06-29
Fix a key not found issue in survey in multi-process environment

isra-fel:
See comment

### DIFF
--- a/src/Share/Survey/SurveyHelper.cs
+++ b/src/Share/Survey/SurveyHelper.cs
@@ -88,12 +88,9 @@ namespace Microsoft.Azure.PowerShell.Common.Share.Survey
 
             int majorVersion = moduleVersion.Major;
 
-if (Modules.Count == 0 && !ReadFromStream())
-{
-  return false
-}
+            if (Modules.Count == 0 && !ReadFromStream())
             {
-                if (!ReadFromStream()) { return false; }
+                return false;
             }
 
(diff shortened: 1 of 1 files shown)


src/Share/Survey/SurveyHelper.cs
```diff
@@ -90,7 +90,7 @@ public bool ShouldPropmtSurvey(string moduleName, Version moduleVersion)
 
             if (Modules.Count == 0)
             {
-                ReadFromStream();
+                if (!ReadFromStream()) { return false; }
```
isra-fel, line 93: Previously, if ReadFromStream returned false (this may happen in multi-process), the Modules dictionary would be empty, and ShouldModulePrompt would throw a key-not-found exception when trying to access the dictionary.

src/Share/Survey/SurveyHelper.cs
```diff
@@ -90,7 +90,7 @@ public bool ShouldPropmtSurvey(string moduleName, Version moduleVersion)
 
             if (Modules.Count == 0)
```
VeryEarly, line 91: NIT
```suggestion
if (Modules.Count == 0 && !ReadFromStream())
{
  return false
}
```

VeryEarly (approved):

BethanyZhou (approved):
LGTM

PR #348, opened 2022-09-07
Upgrade ApplicationInsights 2.13.1

dingmeng-xue:
This pull request upgrade ApplicationInsights to 2.13.1. It is the version supported by PowerShell 7.0.0

Another change is to make flush method to async because sync method flush impacts customer's experience when network to AI endpoint is very slow.  

Remove private variable `FlushTimeoutInMilli`


dingmeng-xue:
Please don't merge this pull request till I verify that in azure-powershell repo.

vidai-msft (approved):
Looks good!

src/Common/MetricHelper.cs
```diff
@@ -481,17 +482,27 @@ public void FlushMetric()
...
+            foreach (TelemetryClient client in TelemetryClients)
             {
-                foreach (TelemetryClient client in TelemetryClients)
+                if(cancellationToken != CancellationToken.None && cancellationToken.IsCancellationRequested)
+                {
+                    return Task.FromCanceled(cancellationToken);
+                }
+                try
                 {
                     client.Flush();
```
isra-fel, line 498: We should call the async version of the API, otherwise it's still running synchronously.
```csharp
client.FlushAsync();
```

dingmeng-xue: Async issue is fixed. Please review.

isra-fel:
As we are in the middle of ADO migration, CI needs to be manually triggered:
https://dev.azure.com/azclitools/internal/_build?definitionId=148

src/Common/MetricHelper.cs
```diff
@@ -481,17 +481,25 @@ public void FlushMetric()
                 return;
             }
 
-            try
+            FlushMetricAsync(TelemetryClients);
```
isra-fel, line 484: A warning will be throwed when we don't `await` an `async` method, the recommended way to suppress the warning is by assigning the returned task to a variable.

> You should consider suppressing the warning only if you're sure that you don't want to wait for the asynchronous call to complete and that the called method won't raise any exceptions. In that case, you can suppress the warning by assigning the task result of the call to a variable.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs4014

```suggestion
            Task flushInBackground = FlushMetricAsync(TelemetryClients);
```

isra-fel (changes requested):

src/Common/MetricHelper.cs
dingmeng-xue: Updated.

isra-fel (approved):

vidai-msft (approved):
Looks good!

PR #353, opened 2022-09-25
Deprecate Azure German Cloud

dingmeng-xue:
According to announcement https://www.microsoft.com/en-us/cloud-platform/germany-cloud-regions, this pull request removes German Cloud from environment list. Related public variables have added obsolete attribute.

### DIFF
--- a/src/Authentication.Abstractions/AzureEnvironmentConstants.cs
+++ b/src/Authentication.Abstractions/AzureEnvironmentConstants.cs
@@ -12,6 +12,8 @@
 // limitations under the License.
 // ----------------------------------------------------------------------------------
 
+using System;
+
 namespace Microsoft.Azure.Commands.Common.Authentication.Abstractions
 {
     /// <summary>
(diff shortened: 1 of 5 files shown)


wyunchi-ms (approved):

isra-fel (approved):

src/Authentication.Abstractions/AzureEnvironmentConstants.cs
```diff
@@ -86,6 +95,7 @@ public static class AzureEnvironmentConstants
         public const string AzureActiveDirectoryEndpoint = "https://login.microsoftonline.com/";
         public const string ChinaActiveDirectoryEndpoint = "https://login.chinacloudapi.cn/";
         public const string USGovernmentActiveDirectoryEndpoint = "https://login.microsoftonline.us/";
+        [Obsolete("Microsoft Cloud Germany was closed on October 29th, 2021.")]
```
jiasli, line 98: Why is the constant marked as `Obsolete` instead of being removed if they are not accessible anymore (https://github.com/Azure/azure-cli/issues/23775)?

isra-fel: This repo contains code that is depended by all the azure powershell modules. It's a general practice to maintain backwards compatibility here. Removing this property would be a binary breaking change.

PR #354, opened 2022-09-26
Add public method to support old version of predictor

Nickcandy:


dingmeng-xue:
All public method cannot be changed. So, you need to keep both there. Add obsolete attribute to ShouldPropmtAzSurvey() method.

dingmeng-xue (approved):

VeryEarly (approved):

PR #361, opened 2022-10-21
Added interface to define the function to log raw data for test coverage and added more info for AzurePSQosEvent

vidai-msft:


src/Common/AzurePSCmdlet.cs
```diff
@@ -747,18 +750,14 @@ protected void LogQosEvent()
                 return;
             }
 
+            _qosEvent.ParameterSetName = this.ParameterSetName;
```
isra-fel, line 753: This is already done by https://github.com/Azure/azure-powershell-common/blob/b6a77a31d2c6bcf2036fdb99e9bc5aef26c6fd72/src/Common/AzurePSCmdlet.cs#L648
Why do it again?

src/Authentication.Abstractions/Interfaces/IPSCmdletAction.cs
```diff
@@ -0,0 +1,23 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// ----------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Azure.Commands.Common.Authentication.Abstractions
+{
+    public interface IPSCmdletAction
```
isra-fel, line 19: The name `IPSCmdletAction` looks too general to me. Can we be more specific so it's more obvious about its function? For example `ITestCoverageUtility` or `ITestCoverageHelper`

src/Authentication.Abstractions/Interfaces/IPSCmdletAction.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        void LogTestCoverageRawData(string moduleName, string commandName, string parameterSetName, string parameters, string sourceScript, int scriptLineNumber, bool isSuccess);
```
isra-fel, line 21: This method has a lot of parameters, and all of them actually come from QosEvent. Can it have only one parameter?
```suggestion
        void LogTestCoverageRawData(QosEvent e);
        // or preferable using a struct:
        void LogTestCoverageRawData(TestCoverageRawData d);
        struct TestCoverageRawData
        {
          string ModuleName;
          ...
        }
```

isra-fel (commented):
Please see the inline comments. Also, can we move the interface to common.csproj? Because it's not really related to authentication and it's only used in common.csproj.

src/Authentication.Abstractions/Interfaces/IPSCmdletAction.cs
isra-fel: Please add inline docs as well

src/Common/AzurePSCmdlet.cs
vidai-msft: AutoRest cmdlet has issue for the value of ParameterSetName. That is why we need to re-assign.

src/Authentication.Abstractions/Interfaces/IPSCmdletAction.cs
vidai-msft: The original reason is AzurePSQosEvent is defined in the Common project and Abstraction project does not have reference of the Common. Otherwise, it requires to add more project references. I will try to move the interface to the Common project and then this can be resolved.

src/Authentication.Abstractions/Interfaces/IPSCmdletAction.cs
vidai-msft: Moved IPSCmdletAction from Abstraction to Common and renamed to ITestCoverage.

src/Authentication.Abstractions/Interfaces/IPSCmdletAction.cs
vidai-msft: Moved and replaced the parameters with AzurePSQosEvent

isra-fel (approved):

PR #369, opened 2022-12-12
Allow doing stuff after config is applied

isra-fel:
Extends `ConfigDefinition` type to allow configs to define actions *after* a new value is applied. (Adding a virtual method to an abstract class is not a binary breaking change.)
This is part of the work to support configuring login with WAM, as the `AuthenticatorBuilder` needs to be reset *after* the WAM config is updated.


isra-fel:
This is no longer necessary as we changed the design of the WAM config.
Now that the WAM authenticator is always added to the authenticator chain, so there's no need to do anything after the config is applied. (see https://github.com/Azure/azure-powershell/pull/17466)
I'll close this PR but keep the branch until we need it.

PR #374, opened 2023-02-22
Support ArmMetadata with API version 2022-09-01

msJinLei:
ArmMetadata with API version 2022-09-01 has changed the response format from array to single object. Also properties in the metadata are also changed.
image

- Update the related code to support new format of metadata
- Use the properties in the metadata to update the endpoints of MicrosoftGraph, Attestation and SynapseAnalytics endpoints
- Fix the target framework issue of test projects

### DIFF
--- a/src/Authentication.Abstractions/AzureEnvironment.cs
+++ b/src/Authentication.Abstractions/AzureEnvironment.cs
@@ -33,6 +33,7 @@ namespace Microsoft.Azure.Commands.Common.Authentication.Abstractions
     {
         private const string ArmMetadataEnvVariable = "ARM_CLOUD_METADATA_URL";
 
+        private const string DefaultArmMetaDataEndpoint = "https://management.azure.com/metadata/endpoints?api-version=2022-09-01";
         private const string DisableArmMetaDataEndpoint = "DISABLED";
 
         internal static IDictionary<string, AzureEnvironment> InitializeBuiltInEnvironments(string armMetadataRequestUri, Action<string> debugLogger = null, Action<string> warningLogger = null, IHttpOperations httpOperations = null)
@@ -56,13 +57,21 @@ namespace Microsoft.Azure.Commands.Common.Authentication.Abstractions
                 if (!string.IsNullOrEmpty(armMetadataRequestUri))
                 {
                     debugLogger?.Invoke($"Discover environments via {armMetadataRequestUri}");
-                    var list = InitializeEnvironmentsFromArm(armMetadataRequestUri, httpOperations).ConfigureAwait(false).GetAwaiter().GetResult();
-                    foreach (var metadata in list)
+                    try
                     {
-                        var env = MapArmToAzureEnvironment(metadata);
-                        env.Type = TypeDiscovered;
-                        armAzureEnvironments[env.Name] = env;
-                        debugLogger?.Invoke($"Added discovered environment {env.Name}");
+                        var list = InitializeEnvironmentsFromArm(armMetadataRequestUri, httpOperations).ConfigureAwait(false).GetAwaiter().GetResult();
+                        foreach (var metadata in list)
+                        {
+                            var env = MapArmToAzureEnvironment(metadata);
+                            env.Type = TypeDiscovered;
+                            armAzureEnvironments[env.Name] = env;
+                            debugLogger?.Invoke($"Added discovered environment {env.Name}");
+                        }
+                    }
+                    catch (Exception e)
+                    {
+                        warningLogger?.Invoke($"Cannot discover environments from Arm with Uri {armMetadataRequestUri}");
+                        warningLogger?.Invoke(e.StackTrace);
                     }
                 }
             }
@@ -191,7 +200,8 @@ namespace Microsoft.Azure.Commands.Common.Authentication.Abstractions
             if (!string.IsNullOrEmpty(armMetadata.MicrosoftGraphResourceId))
             {
                 azureEnvironment.SetProperty(ExtendedEndpoint.MicrosoftGraphEndpointResourceId, armMetadata.MicrosoftGraphResourceId);
-                //Default ARM endpoint doens't provide MicrosoftGraphUrl. Keep it here just in case.
+                // ARM endpoint only gives us graph resource ID (with ending slash "/"),
+                // we assume the Url (endpoint to where we send requests) equals the resource ID without the slash
                 if (armMetadata.MicrosoftGraphResourceId.EndsWith("/"))
                 {
                     azureEnvironment.SetProperty(ExtendedEndpoint.MicrosoftGraphUrl,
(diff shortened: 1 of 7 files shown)


BethanyZhou (approved):
LGTM

src/Authentication.Abstractions/AzureEnvironment.cs
```diff
@@ -314,6 +188,37 @@ private static AzureEnvironment MapArmToAzureEnvironment(ArmMetadata armMetadata
                 azureEnvironment.ServiceManagementUrl += "/";
             }
 
+            if (!string.IsNullOrEmpty(armMetadata.MicrosoftGraphResourceId))
+            {
+                azureEnvironment.SetProperty(ExtendedEndpoint.MicrosoftGraphEndpointResourceId, armMetadata.MicrosoftGraphResourceId);
+                //Default ARM endpoint doens't provide MicrosoftGraphUrl. Keep it here just in case.
```
isra-fel, line 194:
```suggestion
                // ARM endpoint only gives us graph resource ID (with ending slash "/"),
                // we assume the Url (endpoint to where we send requests) equals the resource ID without the slash
```

src/Authentication.Abstractions/AzureEnvironment.cs
```diff
@@ -314,6 +188,37 @@ private static AzureEnvironment MapArmToAzureEnvironment(ArmMetadata armMetadata
...
+            if (!string.IsNullOrEmpty(armMetadata.SynapseAnalyticsResourceId))
+            {
+                azureEnvironment.SetProperty(ExtendedEndpoint.AzureSynapseAnalyticsEndpointSuffix, armMetadata.SynapseAnalyticsResourceId);
+                if (!string.IsNullOrEmpty(armMetadata.Suffixes.SynapseAnalytics))
+                {
+                    azureEnvironment.SetProperty(ExtendedEndpoint.AzureSynapseAnalyticsEndpointSuffix, armMetadata.Suffixes.SynapseAnalytics);
+                }
+            }
+            //Todo:MhsmDns=ManagedHsmServiceEndpointSuffix;AzureManagedHsmServiceEndpointResourceId not provided
+            //Todo:AcrLoginServer=ContainerRegistryEndpointSuffix;ContainerRegistryEndpointResourceId not provided
```
isra-fel, line 220: Can you explain more (in the code) about these todos?
What do we need to do and when can we do it?

src/Authentication.Abstractions/AzureEnvironmentBuiltInSettings.cs
```diff
@@ -0,0 +1,202 @@
...
+using System.Linq;
+using System.Net;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Commands.Common.Authentication.Abstractions
+{
+    /// <summary>
+    /// The build-in endpoints settings of AzureEnvironment.
+    /// </summary>
+    public partial class AzureEnvironment
```
isra-fel, line 30: I like the idea of moving some related code out into a partial class.
Considering naming the file `AzureEnvironment.BuiltIn.cs` to match the pattern we've been using for partial classes in this repo.

isra-fel (changes requested):
Please see my inline comments.
ALso make sure to test with previous version of Az.Resources (MSgraph cmdlets).
Thanks!

src/Authentication.Abstractions/AzureEnvironment.cs
```diff
@@ -29,11 +29,10 @@ namespace Microsoft.Azure.Commands.Common.Authentication.Abstractions
     /// location fo service-specific endpoints, and information for bootstrapping authentication
     /// </summary>
     [Serializable]
-    public class AzureEnvironment : IAzureEnvironment, IEquatable<AzureEnvironment>
+    public partial class AzureEnvironment : IAzureEnvironment, IEquatable<AzureEnvironment>
     {
         private const string ArmMetadataEnvVariable = "ARM_CLOUD_METADATA_URL";
 
-        private const string DefaultArmMetaDataEndpoint = "http://discover.azure.com/";
```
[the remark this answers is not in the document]
msJinLei: The default point is no longer accessible since years ago. However our product spends time to try to access it every-time we import Az.Accounts and so just remove it.

src/Authentication.Abstractions/AzureEnvironment.cs
```diff
@@ -200,65 +118,21 @@ private static async Task<List<ArmMetadata>> InitializeEnvironmentsFromArm(strin
...
-                azureEnvironments[EnvironmentName.AzureUSGovernment].SetProperty(ExtendedEndpoint.ManagedHsmServiceEndpointResourceId, AzureEnvironmentConstants.USGovernmeneManagedHsmServiceEndpointResourceId);
-                azureEnvironments[EnvironmentName.AzureUSGovernment].SetProperty(ExtendedEndpoint.ManagedHsmServiceEndpointSuffix, AzureEnvironmentConstants.USGovernmentManagedHsmDnsSuffix);
-                azureEnvironments[EnvironmentName.AzureUSGovernment].SetProperty(ExtendedEndpoint.MicrosoftGraphEndpointResourceId, AzureEnvironmentConstants.USGovernmentMicrosoftGraphEndpointResourceId);
-                azureEnvironments[EnvironmentName.AzureUSGovernment].SetProperty(ExtendedEndpoint.MicrosoftGraphUrl, AzureEnvironmentConstants.USGovernmentMicrosoftGraphUrl);
-                azureEnvironments[EnvironmentName.AzureUSGovernment].SetProperty(ExtendedEndpoint.ContainerRegistryEndpointResourceId, AzureEnvironmentConstants.USGovernmentContainerRegistryEndpointResourceId);
+                var metadata = JsonConvert.DeserializeObject<ArmMetadata>(armMetadataContent);
```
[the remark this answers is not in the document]
msJinLei:
```
            try
            {
                return JsonConvert.DeserializeObject<List<ArmMetadata>>(armMetadataContent);
            }
            catch
            {
                var metadata = JsonConvert.DeserializeObject<ArmMetadata>(armMetadataContent);
                if (metadata!= null)
                {
                    var armMetadataList = new List<ArmMetadata>();
                    armMetadataList.Add(metadata);
                    return armMetadataList;
                }
            }
```

src/Authentication.Abstractions/AzureEnvironment.cs
msJinLei: AzureManagedHsmServiceEndpointResourceId and ManagedHsmServiceEndpointSuffix are currently hard-coded in our product. and Now ManagedHsmServiceEndpointSuffix(MhsmDns) is provided by ArmMetadata. But it doesn't' make sense to just refresh ManagedHsmServiceEndpointSuffix by MhsmDns from ARM. If we want to refresh AzureManagedHsmServiceEndpointResourceId with reference to ManagedHsmServiceEndpointSuffix,  it may require the process of verification as something we do for MicrosoftGraph endpoints.

I think we can do it when ManagedHsmService endpoints and ContainerRegistry endpoints are also required from service team.

src/Authentication.Abstractions/AzureEnvironment.cs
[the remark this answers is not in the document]
isra-fel: We'll keep the default endpoint because we still need it to query the endpoints when `Add-AzEnvironment -AutoDiscover`

src/Authentication.Abstractions/AzureEnvironment.cs
isra-fel: It makes sense. Please put it in code as comments, so we'll understand the context when revist it in the future.

BethanyZhou (dismissed):

src/Authentication.Abstractions/AzureEnvironment.cs
```diff
@@ -44,30 +44,34 @@ public class AzureEnvironment : IAzureEnvironment, IEquatable<AzureEnvironment>
...
+                            var env = MapArmToAzureEnvironment(metadata);
+                            env.Type = TypeDiscovered;
+                            armAzureEnvironments[env.Name] = env;
+                            debugLogger?.Invoke($"Added discovered environment {env.Name}");
+                        }
+                    }
+                    catch (Exception e)
+                    {
+                        warningLogger?.Invoke($"Cannot discover environments from Arm with Uri {armMetadataRequestUri}");
+                        warningLogger?.Invoke(e.StackTrace);
```
isra-fel, line 74: We should be formal with strings visible to customers.
Also it's generally not a good idea to write a lot of warnings, especially when the error is not blocking.
```suggestion
                        warningLogger?.Invoke($"Cannot discover environments from ARM with URI {armMetadataRequestUri}. Falling back to built-in environments.");
                        debugLogger?.Invoke(e.StackTrace);
```

isra-fel:
Since the PR updated code related to data plane endpoints and dns suffix, please test data plane cmdlets.
Please also test older versions of service modules to make sure there is no breaking change.

isra-fel (approved):

BethanyZhou (approved):

PR #395, opened 2023-07-12
Change the default warning message of CmdletPreviewAttribute and allow setting ETA

BethanyZhou:
image

### DIFF
--- a/src/Common/AzurePSCmdlet.cs
+++ b/src/Common/AzurePSCmdlet.cs
@@ -47,7 +47,7 @@ namespace Microsoft.WindowsAzure.Commands.Utilities.Common
         IAzureEventListener _azureEventListener;
         protected static ConcurrentQueue<string> InitializationWarnings { get; set; } = new ConcurrentQueue<string>();
 
-        protected static ConcurrentQueue<string> AnnouncedPreviewMessageCmdlets { get; set; } = new ConcurrentQueue<string>();
+        protected static ConcurrentQueue<string> PromptedPreviewMessageCmdlets { get; set; } = new ConcurrentQueue<string>();
 
         private RecordingTracingInterceptor _httpTracingInterceptor;
         private object lockObject = new object();
@@ -384,11 +384,12 @@ namespace Microsoft.WindowsAzure.Commands.Utilities.Common
                 // Write preview message once for each cmdlet in one session
                 // Preview message may be outputed more than once if a cmdlet is concurrently called
                 // Considering lock affects performance but warning message is no harm
-                if (!AnnouncedPreviewMessageCmdlets.Contains(this.MyInvocation.MyCommand.Name)
+                if (this.MyInvocation?.MyCommand != null
+                    && !PromptedPreviewMessageCmdlets.Contains(this.MyInvocation?.MyCommand?.Name)
                     && PreviewAttributeHelper.ContainsPreviewAttribute(this.GetType(), this.MyInvocation))
                 {
                     PreviewAttributeHelper.ProcessCustomAttributesAtRuntime(this.GetType(), this.MyInvocation, WriteWarning);
-                    AnnouncedPreviewMessageCmdlets.Enqueue(this.MyInvocation.MyCommand.Name);
+                    PromptedPreviewMessageCmdlets.Enqueue(this.MyInvocation.MyCommand.Name);
                 }
             }
         }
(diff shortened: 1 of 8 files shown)


src/Common/AzurePSCmdlet.cs
```diff
@@ -378,7 +380,16 @@ private void WriteBreakingChangeOrPreviewMessage()
                 && configManager.GetConfigValue<bool>(ConfigKeysForCommon.DisplayBreakingChangeWarning))
             {
                 BreakingChangeAttributeHelper.ProcessCustomAttributesAtRuntime(this.GetType(), this.MyInvocation, WriteWarning);
-                PreviewAttributeHelper.ProcessCustomAttributesAtRuntime(this.GetType(), this.MyInvocation, WriteDebug);
+
+                // Write preview message once for each cmdlet in one session
+                // Preview message may be outputed more than once if a cmdlet is concurrently called
+                // Considering lock affects performance but warning message is no harm
+                if (!AnnouncedPreviewMessageCmdlets.Contains(this.MyInvocation.MyCommand.Name)
```
isra-fel, line 387: Better to check MyInvocation and MyCommand are not null, like
https://github.com/Azure/azure-powershell-common/blob/dfe6dba5490467c7855a497b0dfec04c5111f49e/src/Common/AzurePSCmdlet.cs#L683

src/Common/AzurePSCmdlet.cs
```diff
@@ -47,6 +47,8 @@ public abstract class AzurePSCmdlet : PSCmdlet, IDisposable
         IAzureEventListener _azureEventListener;
         protected static ConcurrentQueue<string> InitializationWarnings { get; set; } = new ConcurrentQueue<string>();
 
+        protected static ConcurrentQueue<string> AnnouncedPreviewMessageCmdlets { get; set; } = new ConcurrentQueue<string>();
```
isra-fel, line 50: Announced -> Prompted

src/Common/CustomAttributes/CmdletPreviewAttribute.cs
```diff
@@ -25,19 +26,36 @@ public class CmdletPreviewAttribute : System.Attribute
...
             this._message = Resources.PreviewCmdletMessage;
         }
 
         public CmdletPreviewAttribute(string message)
         {
-            this._message = message;
+            this._message = string.IsNullOrEmpty(message) ? Resources.PreviewCmdletMessage : message;
+        }
+
+        public CmdletPreviewAttribute(string message, string estimatedDateOfGa) : this(message)
```
isra-fel, line 43: Do we already have cases of GA date in azure-powershell repo?

isra-fel, line 43: How do we avoid confusion for dates like "1/2/2023"? Some may interpret it as Jan 2nd, some Feb 1st.

src/Common/CustomAttributes/PreviewAttributeHelper.cs
```diff
@@ -33,16 +34,31 @@ internal class PreviewAttributeHelper
...
 
+        /// <summary>
+        /// Process CmdletExperimentation attribute in runtime
+        /// </summary>
+        /// <param name="type"></param>
+        /// <param name="invocationInfo"></param>
+        /// <param name="writeOutput"></param>
+        public static bool ContainsPreviewAttribute(Type type, InvocationInfo invocationInfo)
+        {
+            return GetAllAttributesInType(type, invocationInfo).Count() > 0;
```
isra-fel, line 59: Just want to double check that although the method is called `GetAllAttributesInType`, it only gets preview attribute and ignores other kinds of attributes.

src/Common/CustomAttributes/CmdletPreviewAttribute.cs
BethanyZhou: Searched whole repo, no such case is found.

src/Common/CustomAttributes/PreviewAttributeHelper.cs
BethanyZhou: Yes, that's true.

src/Common/CustomAttributes/CmdletPreviewAttribute.cs
BethanyZhou: The date assume the date follows "en-US" culture like breaking change attribute, which means dates are written in the month–day–year order.

BethanyZhou: The benefits of using local culture to display estimated GA date is customer are more familiar with their local date. But considering:
- to keep consistent with breaking change attribute, CmdletPreviewAttribute won't display estimatedGaDate with local culture to avoid introduce new confusion.
- Customer may use the output as the input of DateTime, standard date time will avoid parse issue.

isra-fel (approved):

wyunchi-ms (approved):

PR #402, opened 2023-08-07
 Copy Directory from AzureTenant and Type from AzureEnvironment in DeepCopy of Azure Context

msJinLei:
1. Copy `Directory` from `AzureTenant` and `Type` from `AzureEnvironment` in DeepCopy of Azure Context
2. Remove unused codes in `DefaultProfile`


src/Authentication.Abstractions/Extensions/AzureContextExtensions.cs
```diff
@@ -171,6 +171,14 @@ public static IAzureContext DeepCopy(this IAzureContext context)
             deepCopy.Tenant.CopyFrom(context.Tenant);
             deepCopy.Subscription.CopyFrom(context.Subscription);
             deepCopy.Environment.CopyFrom(context.Environment);
+            if (context.Tenant is AzureTenant azureTenant)
+            {
+                ((AzureTenant)deepCopy.Tenant).Directory = azureTenant.Directory;
```
YanaXu, line 176: Can we override the method **CopyFrom** in class **AzureTenantExtensions** and **AzureEnvironmentExtensions** like the following:

- public static void CopyFrom(this AzureTenant tenant, AzureTenant other)
- public static void CopyFrom(this AzureEnvironment environment, AzureEnvironment other)

And remove this copy parts of them in this method?
The point is, override methods will hide the details of copyFrom. If you run copyFrom for IAzureTenant, it goes to IAzureTenant. If you run copyFrom for AzureTenant, it goes to AzureTenant. The caller (like AzureContextExtensions.DeepCopy) should not worry about which one is called.

isra-fel: I like the idea of separating the calls to `CopyFrom` method from its implementation. 

However it will be super confusing to have methods with the same naming, doing subtly different stuff, one takes an interface, the other takes an implementation. Imaging this:

```csharp
// this is confusing
public static void CopyFrom(this IAzureTenant tenant, IAzureTenant other)
public static void CopyFrom(this AzureTenant tenant, AzureTenant other)

// it will be better to rename the latter, so that the difference is more obvious, e.g.
public static void CopyFromAzureTenant(this AzureTenant tenant, AzureTenant other)
```
@msJinLei please add a todo here, whenever we introduce a second property that needs to be copied here, let's extract the logic to a new method. For now, it looks good to me.

isra-fel (changes requested):

src/Authentication.Abstractions/Extensions/AzureContextExtensions.cs
msJinLei: Done

isra-fel (approved):

YanaXu (approved):

PR #407, opened 2023-11-29
support breaking change attribute for dynamic parameters

VeryEarly:
get breaking change attribute for all parameters instead of runtime properties only, in this way breaking change attribute for dynamic parameters are able to be displayed


dolauli (approved):
LGTM

PR #415, opened 2024-04-09
Fix the issue that string is returned instead of object and add one more property in the telemetry

vidai-msft:
This pull request primarily focuses on improving the handling of secret detection in the Azure PowerShell Cmdlet. The changes include a more nuanced approach to secret detection warnings, improved null safety, and adjustments to the `SanitizerTelemetry` class.

The most important changes are:

Improved Secret Detection Warnings:

* [`src/Common/AzurePSCmdlet.cs`]: The `WriteSecretsWarningMessage()` method now checks for `sanitizerInfo.SecretsDetected` before showing a secrets warning. It also differentiates the warning message based on whether any properties were detected.
* `src/Common/Properties/Resources.Designer.cs` and `src/Common/Properties/Resources.resx`: Added a new resource string `DisplaySecretsWarningMessageWithoutProperty` for the case where no specific properties were detected.

Null Safety Improvements:

* [`src/Common/AzurePSCmdlet.cs`]: Null safety has been improved in several places by using the null-conditional operator. This includes the `OutputSanitizer` property and the `SanitizeOutput()` and `InitializeQosEvent()` methods.

Changes to `SanitizerTelemetry`:

* [`src/Common/Sanitizer/SanitizerTelemetry.cs`]: Added a new property `SecretsDetected` to track whether secrets were detected. Also added a new constructor that takes a `showSecretsWarning` parameter. 
* [`src/Common/MetricHelper.cs`]: The `PopulateSanitizerPropertiesFromQos()` method now uses the new `SecretsDetected` property and adjusts the `secrets-detected-properties` value based on whether any properties were detected.


src/Common/Sanitizer/SanitizerTelemetry.cs
```diff
@@ -29,11 +31,17 @@ public class SanitizerTelemetry
 
         public TimeSpan SanitizeDuration { get; set; }
 
+        public SanitizerTelemetry(bool showSecretsWarning)
```
isra-fel, line 34: This is technically a breaking change, because the default constructor (the one without parameters) will be hidden by this. Are we sure about the backward compatibility?

vidai-msft: The change was due to the value in the telemetry may be not consistent with the DisplaySecretsWarning option. Previously the ShowSecretsWarning in the telemetry relied on if the real detection method was invoked. If not, then the value would be the default (false). However, the real value of DisplaySecretsWarning may be true. The initialization of the SanitizerTelemetry just happens in Az.Accounts module (For SDK, Authentication proj; for AutoRest, AzModule). So it won't cause any real breaking change and build error.

isra-fel (approved):
LGTM

Nickcandy (approved):

PR #421, opened 2024-06-12
Add Interface to Send Auth Info to Telemetry

msJinLei:
## Purpose 
To enable Az.Accounts to collect telemetry data for authentication processes. 

## Challenges 
- **Serialized Data Storage**: Authentication processes can be called multiple times in one cmdlet, generating several properties. Only string information can be sent to the telemetry table without changing its existing construction. 
- **Propagating Cmdlet Context**: Authentication methods are called in various ways and multiple layers away from the cmdlet implementation, making it difficult to pass cmdlet invocation information. 
- **Multi-threaded Module and Authentication**: Module import, cmdlet execution, and authentication processes occur in multiple threads, requiring thread-safe caching for middle-level data. 
- **Backward Compatibility**: The interfaces in Azure-Powershell-Common are used by all Az modules, necessitating backward compatibility without changing existing interfaces. 

## Solutions 
- **Serialized Data Storage**: Pick the first authentication call data as “primary” and serialize all other calls as “subsidiary”. The first call is generally the most important authentication process in the cmdlet. 
- **Thread-safe Telemetry Storage**: Create a thread-safe telemetry storage to cache authentication telemetry data, bridging the authentication process and cmdlet, and register the storage as the AzureSession Component. 
- **CmdletContext Propagation**: Define a CmdletContext and propagate it across multiple layers. 
- **New Interfaces**: Create new interfaces rather than changing the existing ones. 

### The info added
- auth-info-head-tokencredentialname
- auth-info-head-authenticationsuccess
- auth-info-head-tokencacheenabled

- Example To update later
image


isra-fel (changes requested):
1. 1* Cmdlet telemetry : N * auth telemetry. Conclusion: data structure => array, but only 1 element. Todo: test if array is OK to be stored in kusto?
2. telemetry key naming convention. Conclusion: use dash. General conclusion: ?
3. Scope. Can we leverage MSAL telemetry? Todo: remove AuthorityUri. Consider adding correlation ID later. Consider enable telemetry of -Environment parameter.
4. leverage Beisi & Lei's design. Conclusion: not much.
5. Codegen => Todo: next sprint. Create task.
6. TokenCacheEnabled => nice to have. Consider risk & effort of testing.

src/Authentication.Abstractions/Interfaces/IAuthenticationFactory.cs
```diff
@@ -94,5 +95,10 @@ public interface IAuthenticationFactory : IHyakAuthenticationFactory
         /// <param name="account">The account to remove credentials for</param>
         /// <param name="tokenCache">The TokenCache to remove credentials from</param>
         void RemoveUser(IAzureAccount account, IAzureTokenCache tokenCache);
+
+        /// <summary>
+        /// Get the information to be recorded in Telemetry
+        /// </summary>
+        IList<IAuthenticationInfo> GetDataForTelemetry();
```
isra-fel, line 102: Updating a core interface can be risky. We must

1. Update all implementation of this interface. I believe there's one in Accounts and another in Accounts' tests.
2. Test collaboration of new Accounts and older version of service modules, especially if they use this interface.

src/Common/AzurePSCmdlet.cs
```diff
@@ -430,6 +430,8 @@ protected override void EndProcessing()
             WriteWarningMessageForVersionUpgrade();
             WriteSecretsWarningMessage();
 
+            _qosEvent.AuthInfo = AzureSession.Instance.AuthenticationFactory.GetDataForTelemetry();
```
isra-fel, line 433: This line is better put in LogQosEvent()

src/Common/MetricHelper.cs
```diff
@@ -292,6 +292,30 @@ public void SetPSHost(PSHost host)
         {
         }
 
+        private static void PopulateAuthenticationInfoFromQos(IEnumerable<IAuthenticationInfo> infos, IDictionary<string, string> eventProperties)
```
isra-fel, line 295: Again I recommend use "Telemetry" instead of "Info" in names. Also "information" isn't countable so never use "infos".

isra-fel, line 295: This method is complex. Can you document how it works to help others understand?

isra-fel (changes requested):

src/Authentication.Abstractions/AuthenticationInfo.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        /// <summary>
+        /// Class name of the TokenCredential, stands for the authentication method
+        /// </summary>
+        public string TokenCredentialName { get; set; }
+
+        /// <summary>
+        /// Authentication process succeed or not.
+        /// </summary>
+        public bool AuthenticationSuccess { get; set; } = false;
```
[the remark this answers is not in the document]
msJinLei: Prefer to set it as negative result biased.

src/Authentication.Abstractions/AuthenticationInfo.cs
```diff
@@ -0,0 +1,66 @@
...
+        [JsonIgnore]
+        [XmlIgnore]
+        public IDictionary<string, string> ExtendedProperties { get; } = new ConcurrentDictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
+        public AuthenticationInfo()
+        {
+            TokenCredentialName = null;
+        }
+
+        public AuthenticationInfo(IAuthenticationInfo other, bool? isSuccess = null)
```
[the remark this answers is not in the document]
msJinLei: > Why is isSuccess `bool?` ? In what case will it be null?

If it is null, use the value in `other`. The case doesn't exist in current code.

src/Common/MetricHelper.cs
msJinLei: > This method is complex. Can you document how it works to help others understand?

The method is changed. Please check the new method.

msJinLei:
No change since last review
- MetricHelper
- AuthTelemetryRecord

msJinLei:
> 1* Cmdlet telemetry : N * auth telemetry. Conclusion: data structure => array, but only 1 element. Todo: test if array is OK to be stored in kusto?

Cannot use array directly, the current solution is to expand the first (most important) authentication telemetry record and serialize the other parts into a json string

> telemetry key naming convention. Conclusion: use dash. General conclusion: ?

> Scope. Can we leverage MSAL telemetry? Todo: remove AuthorityUri. Consider adding correlation ID later. Consider enable telemetry of -Environment parameter.
CorrelationId is added but not way to get id from MSAL, need MSAL team to support

> leverage Beisi & Lei's design. Conclusion: not much.

> Codegen => Todo: next sprint. Create task.
Done, need more test

> TokenCacheEnabled => nice to have. Consider risk & effort of testing.
removed

src/Authentication.Abstractions/AzureCmdletContext.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.Azure.Commands.Common.Authentication.Abstractions.Interfaces;
+
+namespace Microsoft.Azure.Commands.Common.Authentication.Abstractions
+{
+    /// <summary>
+    /// Represents the context for an Azure cmdlet.
+    /// </summary>
+    public class AzureCmdletContext : ICmdletContext
+    {
+        private string cmdletId;
```
BethanyZhou, line 23:
```suggestion
        private string _cmdletId;
```
suggest using '_' as the prefix of private field.

src/Common/MetricHelper.cs
```diff
@@ -292,6 +292,20 @@ public void SetPSHost(PSHost host)
         {
         }
 
+        private static void PopulateAuthenticationPropertiesFromQos(AuthenticationTelemetryData telemetry, IDictionary<string, string> eventProperties)
```
BethanyZhou, line 295:
```suggestion
        private static void PopulateAuthenticationPropertiesFromQos(AzurePSQoSEvent qos, IDictionary<string, string> eventProperties)
```
because the naming of this function is ...FromQos

src/Authentication.Abstractions/Interfaces/IClientFactory.cs
```diff
@@ -38,6 +39,17 @@ public interface IClientFactory: IHyakClientFactory
...
+        /// <summary>
+        /// Create a properly configured AutoRest client using the given target Azure context, named endpoint and cmdlet context called from
+        /// </summary>
+        /// <typeparam name="TClient">The client type to create</typeparam>
+        /// <param name="context">The azure context to target</param>
+        /// <param name="endpoint">The named endpoint the client shoulld target</param>
+        /// <param name="cmdletContext">The cmdlet context called from</param>
+        /// <returns>A client properly authenticated in the given context, properly configured for use with Azure PowerShell, 
+        /// targeting the given named endpoint in the targeted environment</returns>
+        TClient CreateArmClient<TClient>(IAzureContext context, string endpoint, ICmdletContext cmdletContext) where TClient : ServiceClient<TClient>;
```
BethanyZhou, line 51: Could you explain why creating an ArmClient needs cmdletContext?

msJinLei: A lot of partners use Arm Client to handle the authentication. If we want to collect authentication telemetry from their cmdlets, a good way is to add cmdlet context here.
Also the CmdletContext can be treated as the same status as AzureContext. We can pass them both in the same interface.

src/Common/MetricHelper.cs
msJinLei: Good point. Will also discuss with yeming about the naming

src/Authentication.Abstractions/AzureCmdletContext.cs
msJinLei: We have both private data members that start with lower case and "_". Let get a unified decision for it.

src/Common/MetricHelper.cs
```diff
@@ -292,6 +292,27 @@ public void SetPSHost(PSHost host)
...
+        {
+            var record = telemetry?.Head;
+            if (record != null)
+            {
+                eventProperties[$"{AuthTelemetryRecord.AuthTelemetryPropertyHeadPrefix}-{nameof(record.TokenCredentialName).ToLower()}"] = record.TokenCredentialName;
```
msJinLei, line 300: Move out of metric helper

src/Authentication.Abstractions/AuthTelemetryRecord.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the authentication process succeeded or not.
+        /// </summary>
+        public bool AuthenticationSuccess { get; set; } = false;
+
+        /// <summary>
+        /// Gets or sets the correlation ID for the authentication.
+        /// </summary>
+        public string CorrelationId { get; set; }
```
msJinLei, line 41: To record Uti rather then CorrelationId may be more feasible

src/Authentication.Abstractions/AuthenticationTelemetryData.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+    /// <summary>
+    /// Represents the telemetry data for authentication.
+    /// </summary>
+    public class AuthenticationTelemetryData
+    {
+        /// <summary>
+        /// Gets the first record of authentication telemetry data, usually describing the main method of the authentication process.
+        /// </summary>
+        public IAuthTelemetryRecord Head { get; } = null;
```
msJinLei, line 27: How about to use primary and subsequent？

src/Common/MetricHelper.cs
msJinLei: > Move out of metric helper
Let's discuss it later

msJinLei: > Move out of metric helper
Let's discuss it later

src/Authentication.Abstractions/AuthenticationTelemetryData.cs
msJinLei: Done

src/Authentication.Abstractions/Interfaces/IAzurePSCmdletConcurrentVault.cs
```diff
@@ -0,0 +1,111 @@
...
+    /// A thread-safe implementation of <see cref="IAzurePSCmdletDataVault{T}"/> that stores data records in concurrent queues.
+    /// Each queue is associated with a cmdlet context identifier.
+    /// </summary>
+    /// <typeparam name="T">The type of data records stored in the vault.</typeparam>
+    /// <remarks>
+    /// This class provides a thread-safe mechanism for storing and retrieving data records
+    /// associated with cmdlet contexts. It uses <see cref="ConcurrentDictionary{TKey, TValue}"/> and
+    /// <see cref="ConcurrentQueue{T}"/> to ensure thread safety.
+    /// </remarks>
+    public abstract class IAzurePSCmdletConcurrentVault<T> : IAzurePSCmdletDataVault<T>
```
isra-fel, line 33: Correct me if im wrong, do we name abstract classes `I***` like an interface?

src/Authentication.Abstractions/Interfaces/IAzurePSCmdletDataVault.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        /// <summary>
+        /// Gets the count of empty cmdlet contexts in the data vault.
+        /// </summary>
+        int EmptyCmdletContextCount { get; }
+
+        /// <summary>
+        /// Gets the count of key not found occurrences in the telemetry data.
+        /// </summary>
+        int KeyNotFoundCount { get; }
```
isra-fel, line 53: I feel like these properties are too specific (low-level) for the purpose of this interface. How about exposing the keys by a more flexible method like `IEnumerable<ICmdletContext> GetContexts()` (to be renamed)?

src/Authentication.Abstractions/Interfaces/ICmdletContext.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        /// <summary>
+        /// Gets the ID of the cmdlet.
+        /// </summary>
+        string CmdletId { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether the cmdlet context is valid.
+        /// </summary>
+        bool IsValid { get; }
```
isra-fel, line 29: Please help me understand why we need `IsValid`

src/Common/AzurePSCmdlet.cs
```diff
@@ -880,6 +883,13 @@ protected void LogQosEvent()
             _qosEvent.ParameterSetName = this.ParameterSetName;
             _qosEvent.FinishQosEvent();
 
+            if (!AzureSession.Instance.TryGetComponent(AuthenticationTelemetry.Name, out AuthenticationTelemetry authenticationTelemetry))
+            {
+                throw new NullReferenceException(Resources.AuthenticationTelemetryNotRegistered);
```
isra-fel, line 888: 1. `InvalidOperationException` may be a better choice
2. If cx disables telemetry, will this exception be thrown?

src/Authentication.Abstractions/AuthTelemetryRecord.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            TokenCredentialName = null;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AuthTelemetryRecord"/> class based on another instance of <see cref="IAuthTelemetryRecord"/>.
+        /// </summary>
+        /// <param name="other">The other instance of <see cref="IAuthTelemetryRecord"/>.</param>
+        /// <param name="isSuccess">A value indicating whether the authentication was successful or not.</param>
+        public AuthTelemetryRecord(IAuthTelemetryRecord other, bool? isSuccess = null)
```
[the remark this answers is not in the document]
msJinLei: To do the deepcopy

src/Authentication.Abstractions/Interfaces/IAzurePSCmdletConcurrentVault.cs
msJinLei: Done

src/Authentication.Abstractions/Interfaces/ICmdletContext.cs
msJinLei: > Please help me understand why we need `IsValid`

It is to check whether the cmdletcontext is valid, in current implementation, whether it contains a valid cmdletId

src/Authentication.Abstractions/Interfaces/IAzurePSCmdletDataVault.cs
msJinLei: > I feel like these properties are too specific (low-level) for the purpose of this interface. How about exposing the keys by a more flexible method like `IEnumerable<ICmdletContext> GetContexts()` (to be renamed)?

EmptyCmdletContextCount , KeyNotFoundCount  are used to record the errors during telemetry collection, which may be useful when do the trouble-shooting. Do you propose to use a data struct to store these properties?

src/Authentication.Abstractions/AuthTelemetryRecord.cs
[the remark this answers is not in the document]
msJinLei: In most cases, all the AuthTelemetryRecord are almost the same except success or not and so we can use deepcopy to create a new record

src/Common/AzurePSCmdlet.cs
msJinLei: I have checked existing codes. When `IsDataCollectionAllowed()` returns false, the data collection related modules are still registered but the telemetry data is not sent. 
Specifically, in the function you refers to (`LogQosEvent`), all the codes are no need executing when `IsDataCollectionAllowed()` returns false. But I am not sure whether there is a side effect to skip the operations in this function when data collection is not allowed. So I keep the existing null reference check to respect the current implementation.

src/Authentication.Abstractions/AuthTelemetryRecord.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        /// <summary>
+        /// Represents the prefix of properties of the first record of authentication telemetry record.
+        /// </summary>
+        public const string AuthTelemetryPropertyPrimaryPrefix = "auth-info-primary";
+
+        /// <summary>
+        /// Represents the key of the left records of authentication telemetry.
+        /// </summary>
+        public const string AuthTelemetryPropertySubsidiaryKey = "auth-info-subsidiary";
```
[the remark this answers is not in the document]
msJinLei: Done

src/Authentication.Abstractions/Interfaces/IAzurePSCmdletDataVault.cs
msJinLei: Discussed with @isra-fel, decide to remove these variables and record debug log with reference to return value of related methods.

isra-fel (dismissed):
LGTM

NoriZC (approved):

YanaXu (approved):

NoriZC (approved):

PR #422, opened 2024-06-18
Change ConfigMetrics from Array to Dictionary

BethanyZhou:
Change ConfigMetrics from Array to Dictionary, two benefits:

- We could check if this config is added in o(1) complexity. Previous design is o(n). A config could be read many times. In worst cases, List<ConfigMetrics> ConfigMetrics may contain same config info and exceed space limit
- We can define nickname for config. Take LoginExperienceV2 for instance, we can add 
ConfigMetrics.Add(ConfigKeys.LoginExperienceV2, new ConfigMetrics(ConfigKeys.LoginExperienceV2, "LoginExperienceVTwo", "On")) for telemetry as our database can't recognize number value in telemetry field.
image

-- no breaking change is introduced as ConfigMetrics is never used yey.

### DIFF
--- a/src/Common/MetricHelper.cs
+++ b/src/Common/MetricHelper.cs
@@ -508,7 +508,7 @@ namespace Microsoft.WindowsAzure.Commands.Common
             {
                 foreach (var configMetric in qos.ConfigMetrics)
                 {
-                    eventProperties[configMetric.ConfigKey] = configMetric.ConfigValue;
+                    eventProperties[configMetric.Value.ConfigKey] = configMetric.Value.ConfigValue;
                 }
             }
         }
@@ -670,7 +670,7 @@ public class AzurePSQoSEvent
     public string ParameterSetName { get; set; }
     public string InvocationName { get; set; }
 
-    public List<ConfigMetrics> ConfigMetrics { get; private set; } 
+    public Dictionary<string, ConfigMetrics> ConfigMetrics { get; private set; } 
 
     public Dictionary<string, string> CustomProperties { get; private set; }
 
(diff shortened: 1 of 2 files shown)


src/Common/MetricHelper.cs
```diff
@@ -508,7 +508,7 @@ private void PopulateConfigMetricsFromQos(AzurePSQoSEvent qos, IDictionary<strin
             {
                 foreach (var configMetric in qos.ConfigMetrics)
                 {
-                    eventProperties[configMetric.ConfigKey] = configMetric.ConfigValue;
+                    eventProperties[configMetric.Value.ConfigKey] = configMetric.Value.ConfigValue;
```
isra-fel, line 511: It can be confusing that ConfigKey doesn't necessary mean the ConfigKey. Consider adding a property to ConfigMetrics to represent the telemetry key.

src/Common/MetricHelper.cs
```diff
@@ -508,7 +508,7 @@ private void PopulateConfigMetricsFromQos(AzurePSQoSEvent qos, IDictionary<strin
             {
                 foreach (var configMetric in qos.ConfigMetrics)
                 {
-                    eventProperties[configMetric.ConfigKey] = configMetric.ConfigValue;
+                    eventProperties[configMetric.Value.TelemetryKey] = configMetric.Value.ConfigValue;
```
isra-fel, line 511: I think using the real config key as the key of eventProperties will simplify the code in Accounts. But you can decide

isra-fel (approved):

PR #425, opened 2024-07-16
Add new properties in the IOutputSanitizer

vidai-msft:
Add IgnoredModules and IgnoredCmdlets to bypass unnecessary modules and/or cmdlets for secrets detection.

### DIFF
--- a/src/Common/Sanitizer/IOutputSanitizer.cs
+++ b/src/Common/Sanitizer/IOutputSanitizer.cs
@@ -12,11 +12,17 @@
 // limitations under the License.
 // ----------------------------------------------------------------------------------
 
+using System.Collections.Generic;
+
 namespace Microsoft.WindowsAzure.Commands.Common.Sanitizer
 {
     public interface IOutputSanitizer
     {
         bool RequireSecretsDetection { get; }
+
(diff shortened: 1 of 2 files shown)


src/Common/Sanitizer/IOutputSanitizer.cs
```diff
@@ -12,12 +12,18 @@
...
 
+using System.Collections.Generic;
+
 namespace Microsoft.WindowsAzure.Commands.Common.Sanitizer
 {
     public interface IOutputSanitizer
     {
         bool RequireSecretsDetection { get; }
 
+        IEnumerable<string> IgnoredModules { get; }
```
VeryEarly, line 23: why IEnumerable not HashSet?

vidai-msft: Just want to keep the interface definition more flexible and enumeration is the necessary feature of the collection. In the implementation, I will use string array instead as IgnoredModules and IgnoredCmdlets are both small lists while HashSet is efficient for large list and looks like a bit heavy for this scenario.

isra-fel (approved):
LGTM

VeryEarly (approved):

PR #426, opened 2024-08-01
Remove PSStyle if terminal does not support escape sequence

lijinpei2008:
- Fix issue: https://github.com/Azure/azure-powershell/issues/24556

- Group in Azure-PowerShell PR: https://github.com/Azure/azure-powershell/pull/25735

- Current effect
image

- Changed effect
image

### DIFF
--- a/src/Common/AzurePSCmdlet.cs
+++ b/src/Common/AzurePSCmdlet.cs
@@ -334,11 +334,21 @@ namespace Microsoft.WindowsAzure.Commands.Utilities.Common
             AzureSession.Instance.ClientFactory.RemoveUserAgent(ModuleName);
             AzureSession.Instance.ClientFactory.RemoveHandler(typeof(CmdletInfoHandler));
         }
+
+        protected virtual void InitializePSStyle()
+        {
+            if (null == PSStyle.SupportsVirtualTerminal)
+            {
+                PSStyle.SupportsVirtualTerminal = Host.UI.SupportsVirtualTerminal;
+            }
+        }
+
         /// <summary>
         /// Cmdlet begin process. Write to logs, setup Http Tracing and initialize profile
         /// </summary>
         protected override void BeginProcessing()
         {
+            InitializePSStyle();
             FlushInitializationWarnings();
             SessionState = base.SessionState;
             var profile = _dataCollectionProfile;
--- a/src/Common/AzurePSCmdlet.cs
+++ b/src/Common/AzurePSCmdlet.cs
@@ -335,20 +335,11 @@ namespace Microsoft.WindowsAzure.Commands.Utilities.Common
             AzureSession.Instance.ClientFactory.RemoveHandler(typeof(CmdletInfoHandler));
         }
 
-        protected virtual void InitializePSStyle()
-        {
-            if (null == PSStyle.SupportsVirtualTerminal)
-            {
-                PSStyle.SupportsVirtualTerminal = Host.UI.SupportsVirtualTerminal;
-            }
-        }
-
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Common/AzurePSCmdlet.cs
```diff
@@ -334,11 +334,21 @@ protected virtual void TearDownHttpClientPipeline()
             AzureSession.Instance.ClientFactory.RemoveUserAgent(ModuleName);
             AzureSession.Instance.ClientFactory.RemoveHandler(typeof(CmdletInfoHandler));
         }
+
+        protected virtual void InitializePSStyle()
+        {
+            if (null == PSStyle.SupportsVirtualTerminal)
+            {
+                PSStyle.SupportsVirtualTerminal = Host.UI.SupportsVirtualTerminal;
+            }
```
isra-fel, line 343: This logic is too specific to be put in AzurePSCmdlet. It should be in PSStyle.

src/Common/ColorAndFormat/PSStyle.cs
```diff
@@ -20,6 +20,8 @@ namespace Microsoft.WindowsAzure.Commands.Common
     /// </summary>
     public sealed class PSStyle
     {
+        public static bool? SupportsVirtualTerminal = null;
+
         /// <summary>
```
isra-fel, line 25: Try something like this

```csharp
public static bool IsEscapeSequenceSupported {
  if (!_isEscapeSequenceSupported.HasValue) { _isEscapeSequenceSupported = Host.UI.SupportsVirtualTerminal; }
  return _isEscapeSequenceSupported.Value;
}
private static bool? _isEscapeSequenceSupported = null;
```

src/Common/AzurePSCmdlet.cs
BethanyZhou: Host.UI.SupportsVirtualTerminal is a property of PSCmdlet, its value is unavailable beyond cmdlet.

src/Common/ColorAndFormat/PSStyle.cs
BethanyZhou: Nice desgin

dolauli (commented):
Do we need to add a change log somewhere for this change.

src/Common/ColorAndFormat/PSStyle.cs
```diff
@@ -20,6 +22,29 @@ namespace Microsoft.WindowsAzure.Commands.Common
...
+                if (_isEscapeSequenceSupported.HasValue)
+                {
+                    return _isEscapeSequenceSupported.Value;
+                }
+                return false;
+            }
+        }
+
+        /// <summary>
+        /// This function should be invoked only in the Az.Accounts.
```
isra-fel, line 40:
```suggestion
        /// Initializes default settings of PSStyle according to the environment.
        /// Note: This function should be invoked only in Az.Accounts.
```

src/Common/ColorAndFormat/PSStyle.cs
```diff
@@ -20,6 +22,29 @@ namespace Microsoft.WindowsAzure.Commands.Common
     /// </summary>
     public sealed class PSStyle
     {
+        private static bool? _isEscapeSequenceSupported = null;
+
+        public static bool IsEscapeSequenceSupported
```
isra-fel, line 27: Any reason to make this public? Otherwise it can be private.

isra-fel (approved):

BethanyZhou:
> Do we need to add a change log somewhere for this change.

@dolauli thanks for asking. No change log is needed for common repo.

dolauli (approved):

PR #427, opened 2024-08-14
set subscription id and tenant id in telemetry if it's null

YanaXu:
It's for `Content-AzAccount`. When it's run, the subscription id and tenant id is not set in context. That's why they are empty in telemetry.
This PR is going to set them before recording the telemetry.

### DIFF
--- a/src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
+++ b/src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
@@ -281,6 +281,22 @@ namespace Microsoft.Azure.Commands.ResourceManager.Common
             return result;
         }
 
+        protected override void EndProcessing()
+        {
+            IAzureContext context;
+            _qosEvent.Uid = "defaultid";
+            if (RequireDefaultContext() && TryGetDefaultContext(out context))
+            {
+                _qosEvent.SubscriptionId = context.Subscription?.Id;
+                _qosEvent.TenantId = context.Tenant?.Id;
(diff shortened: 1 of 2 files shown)


isra-fel (commented):
PR itself looks good.
If we do this, does it make the logic of setting sub id in AzureRMCmdlet.BeginProcessing duplicate, and can we remove that?
Also, this will impact cmdlets that do not inherit AzureRMCmdlet . Please ensure we have tested that scneario. Thanks.

src/ResourceManager/Version2016_09_01/AzureRMCmdlet.cs
```diff
@@ -380,17 +396,7 @@ protected override void InitializeQosEvent()
         {
             base.InitializeQosEvent();
 
-            IAzureContext context;
-            _qosEvent.Uid = "defaultid";
-            if (RequireDefaultContext() && TryGetDefaultContext(out context))
```
isra-fel, line 385: Need more time to check if it's safe to remove `RequireDefaultContext`. Will catch up next week

isra-fel: @YanaXu can you test running some cmdlets that doesn't require a context after clearing the context? They should still work without a context.
Examples are the *DataCollection and ContextAutoSave cmdlets.

isra-fel (approved):
PR looks good if we verified the above cmdlets

msJinLei (approved):

PR #428, opened 2024-08-19
Set UserAgent for ARM telemetry

YanaXu:
This PR is to set UserAgent for ARM request in case of SDK-based cmdlets.
And it fixes the bug of dumplicated CommandName and ParameterSetName.

### DIFF
--- a/src/Common/AzurePSCmdlet.cs
+++ b/src/Common/AzurePSCmdlet.cs
@@ -322,6 +322,17 @@ namespace Microsoft.WindowsAzure.Commands.Utilities.Common
             AzureSession.Instance.ClientFactory.AddUserAgent("AzurePowershell", string.Format("v{0}", AzVersion));
             AzureSession.Instance.ClientFactory.AddUserAgent(PSVERSION, string.Format("v{0}", PowerShellVersion));
             AzureSession.Instance.ClientFactory.AddUserAgent(ModuleName, this.ModuleVersion);
+            try {
+                string hostEnv = AzurePSCmdlet.getEnvUserAgent();
+                if (!String.IsNullOrWhiteSpace(hostEnv)) 
+                {
+                    AzureSession.Instance.ClientFactory.AddUserAgent(hostEnv);
+                }
+            } 
+            catch (Exception) 
+            {
+                // ignore if it failed.
+            }
 
             AzureSession.Instance.ClientFactory.AddHandler(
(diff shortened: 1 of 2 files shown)


src/Common/AzurePSCmdlet.cs
```diff
@@ -322,6 +322,17 @@ protected virtual void SetupHttpClientPipeline()
...
             AzureSession.Instance.ClientFactory.AddUserAgent(PSVERSION, string.Format("v{0}", PowerShellVersion));
             AzureSession.Instance.ClientFactory.AddUserAgent(ModuleName, this.ModuleVersion);
+            try {
+                string hostEnv = AzurePSCmdlet.getEnvUserAgent();
+                if (!String.IsNullOrWhiteSpace(hostEnv)) 
+                {
+                    AzureSession.Instance.ClientFactory.AddUserAgent(hostEnv);
+                }
+            } 
+            catch (Exception)
```
isra-fel, line 332: If we are try-catch `AddUserAgent()`, does it make sense to wrap all `AddUserAgent()` calls in it?

isra-fel (approved):
Approve with minor comments

dolauli (approved):

src/Common/AzurePSCmdlet.cs
YanaXu: > If we are try-catch `AddUserAgent()`, does it make sense to wrap all `AddUserAgent()` calls in it?

Here is `AzureSession.Instance.ClientFactory.AddUserAgent(hostEnv);`. Below is `ClientFactory.RemoveUserAgent(hostEnv);`. The Client part is `UserAgent += string.Format(" {0}", hostEnv);` I can't merge them together.

src/Common/AzurePSCmdlet.cs
```diff
@@ -331,6 +342,18 @@ protected virtual void SetupHttpClientPipeline()
...
         {
+            try
+            {
+                string hostEnv = AzurePSCmdlet.getEnvUserAgent();
+                if (!String.IsNullOrWhiteSpace(hostEnv))
+                {
+                    AzureSession.Instance.ClientFactory.RemoveUserAgent(hostEnv);
+                }
+            }
+            catch (Exception)
```
YanaXu, line 353: I add a catch here just in case the env var is not OK for setting UserAgent, we'll ignore it.

PR #432, opened 2024-10-18
Implement ToInvariantString for DateTime Object

msJinLei:


src/Common/Utilities/InvariantStringConverter.cs
```diff
@@ -0,0 +1,32 @@
...
+// limitations under the License.
+// ----------------------------------------------------------------------------------
+using System;
+using System.Globalization;
+
+namespace Microsoft.Azure.Commands.Common
+{
+    public static  class InvariantStringConverter
+    {
+        private const string InvariantDateTimeFormat = "M/d/yyyy h:mm:ss tt";
```
isra-fel, line 21: Why this specific format?

isra-fel (changes requested):

isra-fel:
As we discussed, there's no "standard" or "always correct" way to serialize a datetime in the request, so let's not provide this helper to prevent it from being abused.

PR #435, opened 2024-11-08
Fix config scope not working

isra-fel:
This pull request includes updates to the `src/Common/AzurePSCmdlet.cs` file to enhance the configuration management by passing an additional parameter, `MyInvocation`, to the `GetConfigValue` method. The changes ensure that the method has more context when retrieving configuration values, which is necessary for the `-AppliesTo` parameter to work.


YanaXu (approved):

vidai-msft (approved):
LGTM

PR #436, opened 2024-11-12
fix race condition in dictionary operations

isra-fel:
Fix Azure/azure-powershell#26624

This pull request Improved the `AzureSession.TryGetComponent` method to use `TryGetValue` instead of `ContainsKey` plus `[]` to avoid potential race condition between the two operations.

### DIFF
--- a/src/Authentication.Abstractions/AzureSession.cs
+++ b/src/Authentication.Abstractions/AzureSession.cs
@@ -30,7 +30,8 @@ namespace Microsoft.Azure.Commands.Common.Authentication
         static IAzureSession _instance;
         static bool _initialized = false;
         static ReaderWriterLockSlim sessionLock = new ReaderWriterLockSlim(LockRecursionPolicy.SupportsRecursion);
-        private IDictionary<ComponentKey, object> _componentRegistry = new ConcurrentDictionary<ComponentKey, object>(new ComponentKeyComparer());
+        // explicit typing for the thread-safe API calls to avoid the need for explicit casting
+        private ConcurrentDictionary<ComponentKey, object> _componentRegistry = new ConcurrentDictionary<ComponentKey, object>(new ComponentKeyComparer());
         private event EventHandler<AzureSessionEventArgs> _eventHandler;
 
         /// <summary>
@@ -263,8 +264,7 @@ namespace Microsoft.Azure.Commands.Common.Authentication
                 () =>
                 {
                     var key = new ComponentKey(componentName, typeof(T));
-                    var components = _componentRegistry as ConcurrentDictionary<ComponentKey, object>;
-                    if (components.TryRemove(key, out var component) && component is IAzureSessionListener listener)
+                    if (_componentRegistry.TryRemove(key, out var component) && component is IAzureSessionListener listener)
                     {
                         _eventHandler -= listener.OnEvent;
                     }
--- a/src/Authentication.Abstractions/AzureSession.cs
+++ b/src/Authentication.Abstractions/AzureSession.cs
@@ -89,7 +89,7 @@ namespace Microsoft.Azure.Commands.Common.Authentication
         public string OldProfileFile { get; set; }
 
         /// <summary>
-        /// The directory contianing the ARM ContextContainer
+        /// The directory containing the ARM ContextContainer
         /// </summary>
         public string ARMProfileDirectory { get; set; }
 
@@ -214,13 +214,16 @@ namespace Microsoft.Azure.Commands.Common.Authentication
         public bool TryGetComponent<T>(string componentName, out T component) where T : class
         {
             var key = new ComponentKey(componentName, typeof(T));
-            component = null;
-            if (_componentRegistry.ContainsKey(key))
+            if (_componentRegistry.TryGetValue(key, out var componentObj) && componentObj is T componentT)
             {
-                component = _componentRegistry[key] as T;
+                component = componentT;
+                return true;
             }
-
-            return component != null;
+            else
+            {
+                component = null;
+                return false;
+            }
         }
 
         public void RegisterComponent<T>(string componentName, Func<T> componentInitializer) where T : class
@@ -234,6 +237,7 @@ namespace Microsoft.Azure.Commands.Common.Authentication
                 () =>
                 {
                     var key = new ComponentKey(componentName, typeof(T));
+                    // todo: double check if dictionary operations are thread safe
                     if (!_componentRegistry.ContainsKey(key) || overwrite)
                     {
                         if (_componentRegistry.ContainsKey(key) && overwrite)
@@ -260,6 +264,7 @@ namespace Microsoft.Azure.Commands.Common.Authentication
                 () =>
                 {
                     var key = new ComponentKey(componentName, typeof(T));
+                    // todo: double check if dictionary operations are thread safe
                     if (_componentRegistry.ContainsKey(key))
                     {
                         var component = _componentRegistry[key];
--- a/src/Authentication.Abstractions/AzureSession.cs
+++ b/src/Authentication.Abstractions/AzureSession.cs
@@ -237,17 +237,16 @@ namespace Microsoft.Azure.Commands.Common.Authentication
                 () =>
                 {
                     var key = new ComponentKey(componentName, typeof(T));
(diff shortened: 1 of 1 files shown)


src/Authentication.Abstractions/AzureSession.cs
```diff
@@ -234,6 +237,7 @@ public void RegisterComponent<T>(string componentName, Func<T> componentInitiali
                 () =>
                 {
                     var key = new ComponentKey(componentName, typeof(T));
+                    // todo: double check if dictionary operations are thread safe
```
vidai-msft, line 240: What does this comment here mean? Anything left undone?

VeryEarly: might need to replace `containskey` and assign by index with atomic methods

vidai-msft: Updated the implementation

BethanyZhou (approved):

src/Authentication.Abstractions/AzureSession.cs
```diff
@@ -236,24 +236,27 @@ public void RegisterComponent<T>(string componentName, Func<T> componentInitiali
...
+                        if (!_componentRegistry.TryGetValue(key, out var existingComponent) || overwrite)
                         {
-                            var existed = _componentRegistry[key];
-                            if (existed is IAzureSessionListener existedListener)
+                            if (overwrite && existingComponent is IAzureSessionListener existedListener)
                             {
                                 _eventHandler -= existedListener.OnEvent;
                             }
+                            var component = componentInitializer();
+                            _componentRegistry[key] = component;
```
VeryEarly, line 250: this line might also need atomic method applied

src/Authentication.Abstractions/AzureSession.cs
```diff
@@ -263,16 +266,22 @@ public void UnregisterComponent<T>(string componentName) where T : class
...
-                        if (component is IAzureSessionListener listener)
+                        var key = new ComponentKey(componentName, typeof(T));
+                        if (_componentRegistry.TryGetValue(key, out var component))
                         {
-                            _eventHandler -= listener.OnEvent;
+                            if (component is IAzureSessionListener listener)
+                            {
+                                _eventHandler -= listener.OnEvent;
+                            }
+                            _componentRegistry.Remove(key);
```
VeryEarly, line 279: this line might also need atomic method applied

VeryEarly (changes requested):

src/Authentication.Abstractions/AzureSession.cs
vidai-msft: Used the sessionLock to ensure thread safety.

src/Authentication.Abstractions/AzureSession.cs
vidai-msft: Same as above

src/Authentication.Abstractions/AzureSession.cs
VeryEarly: 1. we are already using concurrentdictionary to achieve thread safe, I don't think it is proper to introduce another mechanism writelock here
2. if we are trying to use writelock, we do not need trygetvalue() either
3. if we are trying to mix atomic method from concurrentdictionary and writelock, the try/finally block should be applied on line 250 only

VeryEarly: not sure how atomic methods interact with locks, can you provide more insights?

VeryEarly (changes requested):

dolauli (approved):

VeryEarly (changes requested):
considering the writelock instance is also used in modify method, I propose do not use writelock in register/unregister component and use atomic methods instead

src/Authentication.Abstractions/AzureSession.cs
vidai-msft: 1. ReadWriteSlimLock has already been applied in this class. ConcurrentDictionary and lock has no conflicts. ConcurrentDictionary only makes sure the dictionary won't be updated at the same time by different threads (achieved by internal lock) while lock ensures only a single thread can execute the block of codes.
Furthermore, the _componentRegistry is defined as `IDictionary` (definition) instead of `ConcurrentDictionary` (implementation). So we are not able to invoke the TryAdd/TryUpdate/TryRemove without explict type cast. Maybe we can consider redefining its type to `ConcurrentDictionary`.
2. `TryGetValue` is highly recommended getting values. Contains + Get is legacy due to performance overhead especially for large objects. In all means, we should use TryGetValue to retrieve values.
3. I believe we should ensure only one thread can register/unregister the component instead of updating the components dictionary only.

src/Authentication.Abstractions/AzureSession.cs
```diff
@@ -234,16 +237,16 @@ public void RegisterComponent<T>(string componentName, Func<T> componentInitiali
                 () =>
                 {
                     var key = new ComponentKey(componentName, typeof(T));
-                    if (!_componentRegistry.ContainsKey(key) || overwrite)
+                    if (!_componentRegistry.ContainsKey(key) || overwrite) // only proceed if key not found or overwrite is true
```
isra-fel, line 240: I can't rewrite `RegisterComponent()` to be atomic because whether the  new value should be set not only depends on if the key exists, but also if overwrite = true. There's not a method of concurrent dictionary that takes two options.

VeryEarly: does it make sense to reduce number of tryget operation?
```
if ( trygetvalue)
{
    if (overwrite)
    {
        //remove event
    }
    else
    {
        return
    }
}
// add event
```

VeryEarly (approved):

src/Authentication.Abstractions/AzureSession.cs
vidai-msft: Will RegisterComponent run in multi-threaded env? If multi-threads execute _componentRegistry.ContainsKey, I think they may register the same component?

src/Authentication.Abstractions/AzureSession.cs
```diff
@@ -260,14 +263,10 @@ public void UnregisterComponent<T>(string componentName) where T : class
                 () =>
                 {
                     var key = new ComponentKey(componentName, typeof(T));
-                    if (_componentRegistry.ContainsKey(key))
+                    var components = _componentRegistry as ConcurrentDictionary<ComponentKey, object>;
```
vidai-msft, line 266: Can we rename the type from IDictionary to ConcurrentDictionary to reduce the implicit cast?

isra-fel: 1. AzureSession.Instance is a singleton shared across threads, so it's intended that all the components are also singletons. It's OK if a component is registered multiple times, because all threads will use the same component after all.
2. Sure. I've updated that.

src/Authentication.Abstractions/AzureSession.cs
```diff
@@ -234,16 +238,16 @@ public void RegisterComponent<T>(string componentName, Func<T> componentInitiali
...
-                            var existed = _componentRegistry[key];
-                            if (existed is IAzureSessionListener existedListener)
-                            {
-                                _eventHandler -= existedListener.OnEvent;
-                            }
+                            _eventHandler -= existedListener.OnEvent;
                         }
+
                         var component = componentInitializer();
                         _componentRegistry[key] = component;
```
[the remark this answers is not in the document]
isra-fel: The scenario here matches [Store a key/value pair in the dictionary unconditionally, and overwrite the value of a key that already exists](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=net-8.0#remarks:~:text=Store%20a%20key/value%20pair%20in%20the%20dictionary%20unconditionally%2C%20and%20overwrite%20the%20value%20of%20a%20key%20that%20already%20exists) so I followed the recommendation to use the setter (`[]`). And since it's unconditional and always overwrites, I figure there's no need to handle the case that it's unsuccessful. So I always remove the handler.

VeryEarly (approved):

vidai-msft (approved):
LGTM

PR #441, opened 2025-03-25
Fix that RemoveUser Doesn't Work When Create Public Client with Broker

msJinLei:
Add a new interface of "RemoveUser" to accept authority as the input for Authority from environment is required to decide whether public client should be created with broker or not.

### DIFF
--- a/src/Authentication.Abstractions/Interfaces/IAuthenticationFactory.cs
+++ b/src/Authentication.Abstractions/Interfaces/IAuthenticationFactory.cs
@@ -141,7 +141,7 @@ namespace Microsoft.Azure.Commands.Common.Authentication.Abstractions
         /// Remove any stored credentials for the given user and the Azure environment used.
         /// </summary>
         /// <param name="account">The account to remove credentials for</param>
-        /// <param name="authority">The Microsoft Entra authority</param>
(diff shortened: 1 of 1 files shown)


src/Authentication.Abstractions/Interfaces/IAuthenticationFactory.cs
```diff
@@ -133,6 +133,15 @@ IAccessToken Authenticate(
...
+        [Obsolete("RemoveUser is deprecated, please use RemoveUser with Azure environment instead.", true)]
         void RemoveUser(IAzureAccount account, IAzureTokenCache tokenCache);
+
+
+        /// <summary>
+        /// Remove any stored credentials for the given user and the Azure environment used.
+        /// </summary>
+        /// <param name="account">The account to remove credentials for</param>
+        /// <param name="authority">The Microsoft Entra authority</param>
+        void RemoveUser(IAzureAccount account, string authority);
```
isra-fel, line 145: Would it make sense if we require the IAzureEnvironment interface rather than just the authority?

msJinLei: Done

isra-fel (approved):

dolauli (approved):

PR #442, opened 2025-03-30
Fix concurrency issue of Component Registery of Azure Session

msJinLei:
### What to fix 
- In the method `RegisterComponent`, the operation that checks key existance ` _componentRegistry.TryGetValue(key, out var existed)`and the operation that updates the value of the key `_componentRegistry[key] = component` are not atomic.  The result is the when you want to register the components with the same key in different threads in no-overwrite mode,  the component is actually overwritten. What you can see is that with the unfixed codes, the result of the test case `TestComponentRegistrationSameComponentNoOverwritten` is unpredictable.

### DIFF
--- a/src/Authentication.Abstractions.Test/AzureSessionTest.cs
+++ b/src/Authentication.Abstractions.Test/AzureSessionTest.cs
@@ -289,10 +289,10 @@ namespace Authentication.Abstractions.Test
             };
 
             // Register components in parallel
-            var tasks = new List<Task>();
+            var tasks = new ConcurrentBag<Task>();
             foreach (var component in testComponents)
             {
-                Task.Run(() => RegisterAndUnregisterComponent(component, componentValue)).ContinueWith(t => tasks.Add(t));
+                Task.Run(() => RegisterAndUnregisterComponent(component, componentValue)).ContinueWith(t => tasks.Add(t), TaskScheduler.Default);
             }
 
             // Wait for all register tasks to complete
--- a/src/Authentication.Abstractions.Test/AzureSessionTest.cs
+++ b/src/Authentication.Abstractions.Test/AzureSessionTest.cs
@@ -213,6 +213,8 @@ namespace Authentication.Abstractions.Test
             Task.WaitAll(tasks.ToArray());
 
             // Verify the results
+            AzureSession.Instance.TryGetComponent(testComponent, out TestComponent component);
+            Assert.Equal(10, component.Size);
             Assert.Single(tasks.Select(t => t.Result["id"]).Distinct());
             var checkList = tasks.Select(t => t.Result["size"]);
             Assert.Equal(10, checkList.Distinct().Count());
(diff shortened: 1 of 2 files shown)


src/Authentication.Abstractions.Test/AzureSessionTest.cs
```diff
@@ -0,0 +1,342 @@
...
+
+            foreach (var task in tasks)
+            {
+                task.Start();
+            }
+
+            Task.WaitAll(tasks.ToArray());
+
+            // Verify the results
+            var results = new int[10];
```
NoriZC, line 216: What's this results used for?

src/Authentication.Abstractions.Test/AzureSessionTest.cs
```diff
@@ -0,0 +1,342 @@
...
+
+            Assert.Single(tasks.Select(t => t.Result["id"]).Distinct());
+            var checkList = tasks.Select(t => t.Result["size"]);
+            Assert.Equal(10, checkList.Distinct().Count());
+            Assert.Equal(10, checkList.Max());
+            Assert.Equal(1, checkList.Min());
+            AzureSession.Instance.ClearComponents();
```
NoriZC, line 223: There are many assertions, could you please help me understand what's the porpers of line 220-line 222? How do these lines verify the correctness?

src/Authentication.Abstractions/AzureSession.cs
```diff
@@ -237,23 +237,31 @@ public void RegisterComponent<T>(string componentName, Func<T> componentInitiali
...
-                        }
+                            if (!overwrite)
+                            {
+                                hasUpdate = false;
+                                return v;
+                            }
+                            else
+                            {
+                                oldComponent = v;
+                                return componentInitializer();
```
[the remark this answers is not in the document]
isra-fel: ...because according to [AddOrUpdate](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=net-9.0#system-collections-concurrent-concurrentdictionary-2-addorupdate(-0-system-func((-0-1))-system-func((-0-1-1)))) either the second or the third Func parameter gets to execute. So there's no need for a cache

isra-fel (dismissed):

src/Authentication.Abstractions.Test/AzureSessionTest.cs
```diff
@@ -0,0 +1,342 @@
...
+            void CheckResults(List<Task<Dictionary<string, int>>> tasks, int id)
+            {
+                var checkList = tasks.Where(t => t.Result["id"] == id);
+                var count = checkList.Count();
+                Assert.Equal(count, checkList.Distinct().Count());
+                if (count > 0)
+                {
+                    Assert.Equal(count, checkList?.Select(t => t.Result["size"])?.Max());
+                }
+            }
```
NoriZC, line 263: Here `CheckResults()` is extracted but only for line 267 usage, and not so easy to understand why. Could you please explain more here?

msJinLei: The results of same component and overwritten case is somewhat unpredictable, similar to unfixed component registration.  Here I just write some rough rules to check the results.
We can divide the results to the different groups. Each group contains the results we can consider to be from a "SameComponentNoOverwritten" case. The size of the group should be equal to the max lengh of clist. That's to say,  clist is appended one by one in the group. But component overwritten happens when clist reaches the max length in the group so that clist reset to empty again.

NoriZC (approved):

isra-fel (approved):

